PR #2, opened 2017-07-14
Iframe api

hristoterezov:


README.md
```diff
@@ -13,26 +13,27 @@ const {
 } = utils;
```
virtuacoplenny: I noticed you changed this import flow in jitsi-meet-electron to use destructuring assignment for RemoteContorl and setupScreenSharingForWindow. Maybe you should do the same here for consistency?

README.md
```diff
@@ -13,26 +13,27 @@ const {
...
 ```Javascript
-setupScreenSharingForWindow(iframe.contentWindow);
+setupScreenSharingForWindow(iframe/* the Jitsi Meet iframe */);
 ```
 
 ## Example
 
 For examples of installation and usage checkout the [Jitsi Meet Electron](https://github.com/jitsi/jitsi-meet-electron) project.
-
 ## Discuss
```
virtuacoplenny: Not related but noting here as a side thought: maybe this discussion section should be removed because it was also removed from the jitsi-meet readme.

package.json
```diff
@@ -36,5 +36,8 @@
     "eslint": ">=3",
     "eslint-plugin-jsdoc": "*",
     "precommit-hook": "3.0.0"
+  },
+  "publishConfig": {
+    "registry": "https://npm-private.atlassian.io"
```
virtuacoplenny: Why is this being added? I'm not against it, mind you, I just don't understand why it's being added because I don't have much context to the build/release flow.

README.md
```diff
@@ -13,26 +13,27 @@ const {
...
+remoteControl.dispose();
 ```
-If Jitsi Meet is loaded in an iframe HTML element which is stored in the `iframe` variable, the code will look like:
+
+NOTE: `dispose` method will be called automatically when the Jitsi Meet iframe unload.
+
+* **Desktop Sharing**
 ```Javascript
-setupScreenSharingForWindow(iframe.contentWindow);
+setupScreenSharingForWindow(iframe/* the Jitsi Meet iframe */);
```
virtuacoplenny: Do you foresee most of the future util calls needing the iframe passed in?

hristoterezov: Not really! 

But for the remote control we need the iframe. So it looks more consistent. 

Also IMHO it looks easier for the users to pass just an HTML element instead of having to select the window of the iframe. 

Also in order to make it work properly with iframe reloads you have to listen for iframe's load event and call the setup function every time. Now this is embedded into the function.

README.md
hristoterezov: Fixed!

README.md
hristoterezov: Removed!

package.json
hristoterezov: It's added by mistake! I removed it! Thanks for noticing!

PR #3, opened 2017-08-04
feat(alwaysontop): Implement

hristoterezov:


alwaysontop/alwaysontop.js
```diff
@@ -0,0 +1,41 @@
+const {
+    api,
+    ondblclick,
+    onload,
+    onbeforeunload
+} = window.alwaysOnTop;
```
virtuacoplenny: Where is window.alwaysOnTop getting set?

alwaysontop/render.js
```diff
@@ -0,0 +1,138 @@
...
+     * Getter for the large video element in Jitsi Meet.
+     *
+     * @returns {HTMLElement|undefined} the large video.
+     */
+    get _jitsiMeetLargeVideo() {
+        const iframe = this._api.getIFrame();
+        if(!iframe || !iframe.contentWindow || !iframe.contentWindow.document) {
+            return;
+        }
+        return iframe.contentWindow.document.getElementById('largeVideo');
```
virtuacoplenny: Is there some way to avoid scraping the dom? I'm worried that if the id were to change for some reason it would not be obvious this broke.

alwaysontop/alwaysontop.js
hristoterezov: alwaysontop/render.js line 84

alwaysontop/alwaysontop.js
```diff
@@ -0,0 +1,41 @@
...
+window.addEventListener('beforeunload', onbeforeunload);
+
+window.addEventListener('dblclick', ondblclick);
+
+document.addEventListener("DOMContentLoaded", () => {
+    onload();
+});
+
+// load all resources from meet
+api._getAlwaysOnTopResources().forEach(src => loadFile(src));
```
hristoterezov: Mainly. Also this way the toolbar will look exactly like the toolbar in the main window. Also this way we are able to reuse our toolbar implementation from jitsi-meet project.

alwaysontop/main.js
```diff
@@ -0,0 +1,79 @@
...
+ */
+function onAlwaysOnTopWindow(
+        event,
+        url,
+        frameName,
+        disposition,
+        options) {
+    if (frameName === 'AlwaysOnTop') {
+        event.preventDefault();
+        event.newGuest = new BrowserWindow(
```
hristoterezov: When the event's default is prevented we are supposed to construct the BrowserWindow on our own and set it to the newGuest prop. Also this way we are able to set the BrowserWindow's options that we need (like always on top, frameless, etc).

Check https://github.com/electron/electron/blob/master/docs/api/window-open.md#using-chromes-windowopen-implementation and https://github.com/electron/electron/blob/master/docs/api/web-contents.md for more info.

alwaysontop/render.js
hristoterezov: Good point! 

Although I can't think of anything else that will be very different. Any ideas?

The important part here is that even if the external resources for the toolbar are not loaded for some reason at least the video will be displayed.

alwaysontop/alwaysontop.js
virtuacoplenny: Is there a guarantee window.alwaysOnTop will be defined by the time alwaysontop.js is running?

hristoterezov: I'm not sure if there is a guarantee but we are setting it right after window.open. I haven't seen it missing tho. I suggest to leave it like that for now because otherwise we'll have to bring some complexity to the scripts which will slow down displaying the video.

README.md
```diff
@@ -26,11 +27,51 @@ remoteControl.dispose();
...
+
+// iframe - the Jitsi Meet iframe
+setupScreenSharingForWindow(iframe);
+```
+
+#### Always On Top
+Displays a small window with the current active speaker video when the main Jitsi Meet window is not focused.
+
+**Requirements**:
+1. Jitsi Meet should be initialed trough our [iframe API](https://github.com/jitsi/jitsi-meet/blob/master/doc/api.md)
```
hristoterezov: Fixed!

PR #11, opened 2017-11-01
Initial implementation of obtaining wifi stats.

damencho:


wifistats/airport.js
```diff
@@ -0,0 +1,40 @@
...
+ * We search for agrCtlRSSI and agrCtlNoise values.
+ * The result is SNR the difference in decibels between the received signal
+ * and the background noise level.
+ *
+ * @param {string} str - the string which is output of the command.
+ * @param callback
+ */
+function parseOutput(str, callback) {
+    try {
+        var lines = str.split('\n');
```
saghul: const!

wifistats/airport.js
```diff
@@ -0,0 +1,40 @@
...
+/**
+ * Parses the output of the {@link cmdLine}.
+ * We search for agrCtlRSSI and agrCtlNoise values.
+ * The result is SNR the difference in decibels between the received signal
+ * and the background noise level.
+ *
+ * @param {string} str - the string which is output of the command.
+ * @param callback
+ */
+function parseOutput(str, callback) {
```
saghul: this function doesn't do any asynchronous work, why not just return the value or null?

wifistats/index.js
```diff
@@ -0,0 +1,117 @@
+const exec    = require('child_process').exec;
+const async   = require('async');
+// The tools
+const airport = require('./airport');
+const iwconfig  = require('./iwconfig');
```
saghul: note that `iwconfig` is deprecated (has been for quite a while) and `ip` is its successor.

damencho: Shit, it has totally different output :( I will look at that. Thanks.

damencho: Maybe using directly /proc/net/wireless will be better

wifistats/airport.js
damencho: The asynchronous was added now :)

wifistats/airport.js
damencho: fixed

wifistats/index.js
damencho: Using /proc/net/wireless now.

damencho:
There is still work in progress around the jitsi-meet and lib-jitsi-meet part and changes will be needed in setupWiFiStats.

wifistats/index.js
```diff
@@ -0,0 +1,129 @@
...
+/**
+ * Setup getWiFiStats to be available in JitsiMeetElectron object.
+ *
+ * @param iframe - the iframe to use attaching the getWiFiStats function.
+ */
+function setupWiFiStats(iframe) {
+    iframe.addEventListener('load', () => {
+        const ctx = iframe.contentWindow;
+        if(typeof ctx.JitsiMeetJS === "undefined")
+            ctx.JitsiMeetJS = {};
```
virtuacoplenny: In what cases do you expect this error to be encountered? Is there a danger of the global getting overridden?

wifistats/index.js
```diff
@@ -0,0 +1,129 @@
...
+ * @param {object} result stats
+ * @returns {Promise}
+ */
+function getWiFiStats() {
+    return new Promise((resolve, reject) => {
+        if (!toolInstance) {
+            initTools(
+                (err, t, result) => {
+                    if (err) {
+                        reject(err);
```
virtuacoplenny: If a promise is rejected, do the immediately following lines still get executed? Would it make a difference?

wifistats/index.js
```diff
@@ -0,0 +1,129 @@
...
+                    }
+                    toolInstance = t;
+                    resolve(result);
+                });
+            return;
+        }
+
+        getStats((error, result) => {
```
virtuacoplenny: I see a mixture of callback-based functions and other that return promises. Is there a reason for not trying to get everything to be one style? The callback pyramid right now is getting a bit deeper than I would prefer.

wifistats/index.js
```diff
@@ -0,0 +1,129 @@
...
+ * global one which we discover to be working.
+ * @param tool - The tool to use querying for wifi stats.
+ * @callback callback delivering results
+ * @param {Error} error if any
+ * @param {object} result stats
+ */
+function getStats(callback, tool) {
+    if (!tool) {
+        if (toolInstance) {
+            tool = toolInstance;
```
virtuacoplenny: Why did you want getStats to set tool? It would surprise me that a function called getStats is setting tool.

wifistats/airport.js
```diff
@@ -0,0 +1,86 @@
...
+                resultObj.addresses = addresses;
+                callback(null, resultObj);
+            } catch (ex) {
+                // cannot get interface address, lets submit whatever we have
+                callback(null, resultObj);
+            }
+        });
+    } catch (ex) {
+        callback(ex, null);
+    }
```
virtuacoplenny: Why the two levels of try/catch? You don't know where a failure will happen and you can't do error handling early?

virtuacoplenny:
Looks like README.md has usage examples for the various public apis. The wifi tools might benefit from such, especially the part about having to call setup.

package.json
```diff
@@ -28,6 +28,7 @@
   "license": "Apache-2.0",
   "gypfile": true,
   "dependencies": {
+    "async": "^2.5.0",
```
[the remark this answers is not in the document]
damencho: No idea.

wifistats/airport.js
```diff
@@ -0,0 +1,86 @@
+const exec = require('child_process').exec;
+
+const cmdLine = '/System/Library/PrivateFrameworks/Apple80211.framework/Versions/Current/Resources/airport -I';
```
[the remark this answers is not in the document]
damencho: Well if it is not working the tool will not be initialized and nothing will be executed later.

wifistats/airport.js
damencho: Will revisit this.

wifistats/index.js
```diff
@@ -0,0 +1,129 @@
+const exec          = require('child_process').exec;
+const async         = require('async');
```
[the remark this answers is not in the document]
damencho: Well the initial implementation was based on a similiar library, and we talked with callstats (where this will be used) to use callback, but later they moved to Promise.
I will try modifying it and make it promise based and will get rid of async.

wifistats/index.js
damencho: This is the initialization on the first call, which founds which tool will be used, checking all operating systems.

wifistats/index.js
damencho: Will take a look, thanks.

wifistats/index.js
damencho: Will change it to promise.

wifistats/index.js
damencho: Well, as not sure when this will be called we do not want to override the namespace if it exists. I borrowed the code from jitsi-meet.

README.md
```diff
@@ -75,6 +75,33 @@ const {
...
+
+In the **main** electron process:
+```Javascript
+const {
+    setupWiFiStats
+} = require("jitsi-meet-electron-utils");
+
+// jitsiMeetWindow - The BrowserWindow instance
+// of the window where Jitsi Meet is loaded.
+setupWiFiStats(jitsiMeetWindow);
```
virtuacoplenny: Is it accurate that setupWiFiStats also needs to be called on the window itself? I'm assuming, possibly incorrectly, that you copied from the AlwaysOnTop above, but that seems to need to be called with the window and the iframe using two separate apis.

wifistats/airport.js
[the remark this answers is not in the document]
virtuacoplenny: Right, I was asking how likely it would succeed. Like do most modern macbooks have this? Older macs?

wifistats/index.js
```diff
@@ -0,0 +1,114 @@
...
+                    }))
+                    .catch(() => {}),
+                getStats(procwireless)
+                    .then(result => resolve({
+                        tool: procwireless,
+                        result
+                    }))
+                    .catch(() => {}),
+                getStats(netsh)
+                    .then(result => resolve({
```
virtuacoplenny: I don't think this will work the way you want it to. resolve applies to the outer promise. If your first getsStats call finishes and calls resolve, other invocations of resolve should have no effect. What you probably want to do is just pass back the object without calling resolve so that success handler in Promise.all().then(handler) receives the object.

wifistats/index.js
```diff
@@ -0,0 +1,114 @@
...
+                        tool: netsh,
+                        result
+                    }))
+                    .catch(() => {})
+            ]
+        ).then(results => {
+                results.forEach(resultEntry => {
+                    if (resultEntry) {
```
virtuacoplenny: Looking at Promise.all, I think you'll get back 3 results, and each can either be an object with a tool or an empty object, and empty objects would make this truthy check pass. You might want to specifically look for something in the object as well, like resultEntry.tool.

Also the forEach will iterate through all the results. If results[0] has a tool, it'll set toolInstance and resolve. If results[2] has a tool, it'll set toolInstance again. Maybe you want to do a results.find() or some other array method or do a for-loop with an early exit.

wifistats/index.js
```diff
@@ -0,0 +1,114 @@
...
+    return new Promise((resolve, reject) => {
+        exec(tool.cmdLine, function (error, str) {
+            if (error) {
+                reject(error);
+                return;
+            }
+
+            tool.parseOutput(str)
+                .then(result => resolve(JSON.stringify(result)))
+                .catch(error => reject(error));
```
virtuacoplenny: It just occurred to me: do you think it's useful to have some debug logging or do you think it wouldn't be useful right now?

wifistats/index.js
```diff
@@ -0,0 +1,114 @@
...
+/**
+ * Queries for wifi stats.
+ * @returns {Promise}
+ */
+function getWiFiStats() {
+    if (!supportWifiStats) {
+        return Promise.reject(new Error('No known wifi stats tool found'));
+    }
+
+    if (!toolInstance) {
```
virtuacoplenny: The jerk in me will tell you to do a truthy check for the positive case first so readers don't have to do the conversion.

wifistats/airport.js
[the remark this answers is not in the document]
damencho: Cannot find information about that. But anyway if it fails the first time will be marked as unsupported and nothing will be executed.

wifistats/index.js
damencho: Nope for the moment.

wifistats/index.js
```diff
@@ -0,0 +1,99 @@
...
+    // check for running OS and use appropriate tool
+    if (os === 'win32') {
+        tool = netsh;
+    } else if (os === 'darwin') {
+        tool = airport;
+    } else if (os === 'linux') {
+        tool = procwireless;
+    }
+
+    return new Promise((resolve, reject) => {
```
virtuacoplenny: If getStats is already returning a promise I don't think you have to wrap this in an additional promise. Instead you can return getStates().then(...).catch(...). Just make sure to return result from the .then and maybe return a Promise.reject(error) from the catch.

wifistats/index.js
virtuacoplenny: I'm more concerned about the global you declared getting overridden by the lib which I assume creates the global JitsiMeetJS as well. That currently does not happen, right? I'm assuming it doesn't happen.

wifistats/airport.js
virtuacoplenny: Still revisiting this? Is it just safer to have a try/catch instead of trying to detect what could fail and acting more "gracefully"?

wifistats/netsh.js
```diff
@@ -0,0 +1,113 @@
...
+            const resultValues = {};
+
+            for (const line of lines) {
+                const res = line.split(":");
+                if (res.length > 1) {
+                    const key = res[0].trim();
+                    resultValues[key] = res[1].trim();
+                }
+            }
+            const iface = resultValues.Name;
```
virtuacoplenny: iface stands for interface?

wifistats/procwireless.js
```diff
@@ -0,0 +1,106 @@
...
+ * @return {Promise}
+ */
+function parseOutput(str) {
+    return new Promise((resolve, reject) => {
+        try {
+            const lines = str.split('\n');
+
+            if (lines.length <= 2) {
+                reject(new Error('No wifi interface'));
+            }
```
virtuacoplenny: Did you want to do an early return here too?

wifistats/procwireless.js
```diff
@@ -0,0 +1,106 @@
...
+
+            // we take the third line
+            const line = lines[2];
+
+            const elements = line.split(':');
+            if (elements.length < 2) {
+                reject(new Error('No wifi interface - wrong format'));
+            }
```
virtuacoplenny: Maybe an early return here too?

wifistats/index.js
```diff
@@ -0,0 +1,99 @@
...
+    let tool;
+
+    // check for running OS and use appropriate tool
+    if (os === 'win32') {
+        tool = netsh;
+    } else if (os === 'darwin') {
+        tool = airport;
+    } else if (os === 'linux') {
+        tool = procwireless;
+    }
```
virtuacoplenny: Is this what replaced the Promise.all that used to exist?

wifistats/netsh.js
damencho: yes

index.js
```diff
@@ -4,10 +4,13 @@ const {
     setupAlwaysOnTopRender,
     setupAlwaysOnTopMain
 } = require('./alwaysontop');
+const { getWiFiStats, setupWiFiStats } = require('./wifistats');
 
 module.exports = {
+    getWiFiStats,
```
virtuacoplenny: Do you still intend to expose getWiFiStats for public use? I only ask because you didn't document it in the README and it seems like setupWiFiStats exposes a way to access getWiFiStats. Maybe the README should also mention how it sets getWiFiStats on a global.

wifistats/index.js
damencho: Yes.

wifistats/index.js
damencho: Yep, it is working. 
Ok, let's discuss this offline tomorrow, cause I'm really confused about this. As I had exposed the function directly in the window, but everybody says that we want to move away from that and expose everything under common namespace and advice me to use this namespace the way we do in jitsi-meet. So your concern is also valid and for jitsi-meet. Maybe I had messed up something and didn't understand.

damencho:
Just pushed a update for your comments.

wifistats/index.js
virtuacoplenny: My bad, I didn't realize you were making separate commits. I could have answered that myself easily but thanks for humoring me.

damencho: I didn't mean to :) sorry

virtuacoplenny:
Looking good to me. I can do some additional passes when the callstats end is more finalized.

wifistats/index.js
virtuacoplenny: I woke up and the first thing that came to mind was how good of a change this was to move away from Promise.all.

virtuacoplenny:
What would you think of adding some unit tests for your changes? It seems to me a lot of your parsing callbacks would be ripe candidates for unit tests because the input values are not recognizable from reading the code and the outputs are predictable. I made a sample implementation here: https://github.com/virtuacoplenny/jitsi-meet-electron-utils/tree/lenny/mocha

damencho:
Hey, @virtuacoplenny I've added some tests, borrowed the code from you :) Does it look ok now?

.eslintrc.js
```diff
@@ -2,7 +2,8 @@ module.exports = {
     'env': {
         'browser': true,
         'commonjs': true,
-        'es6': true
+        'es6': true,
+        "jest": true
```
virtuacoplenny: Maybe this should be "mocha" instead of "jest"?

test/netsh.test.js
```diff
@@ -0,0 +1,94 @@
...
+            const result
+                = netsh._parseIPConfigResults(exampleIPAddressResult1, 'Wi-Fi');
+
+            assert.deepEqual([
+                    '2001:db8:21da:7:713e:a426:d167:37ab',
+                    '157.60.14.11'],
+                result);
+        });
+    });
+    describe('_parseIPAddressResults', () => {
```
virtuacoplenny: describe and it blocks can be nested. So you would only need one describe block that has two it blocks nested in it.

virtuacoplenny:
@damencho, they're beautiful.

virtuacoplenny:
I'm so excited right now. What's great is I can see what the inputs look like. I don't have to imagine what the inputs might be.

wifistats/index.js
```diff
@@ -0,0 +1,96 @@
...
+const netsh         = require('./netsh');
+
+let toolInstance;
+let supportWifiStats = true;
+
+/**
+ * Uses all available tools to query for wifi stats, and the one we found to
+ * work we store in global variable {@link toolInstance}, otherwise we set
+ * supportWifiStats to false.
+ * We call all tools in parallel and when all finishes we check the results.
```
virtuacoplenny: This should be updated now or later.

wifistats/procwireless.js
```diff
@@ -0,0 +1,149 @@
...
+ * @return {Promise}
+ */
+function parseOutput(str) {
+    return new Promise((resolve, reject) => {
+        try {
+            let resultObj;
+            try {
+                resultObj = _parseProcWirelessResults(str);
+            } catch (ex) {
+                reject(ex);
```
virtuacoplenny: Maybe there should be an early return here on error?

PR #14, opened 2017-11-21
fix(alwaysontop): mouse events on windows and linux

hristoterezov:


alwaysontop/render.js
```diff
@@ -95,6 +102,17 @@ class AlwaysOnTop {
             ondblclick: () => {
                 this._closeAlwaysOnTopWindow();
                 this._jitsiMeetElectronWindow.show();
+            },
+            /**
+             * On Windows and Linux if we use the standart drag
```
virtuacoplenny: standart -> standard

PR #15, opened 2017-12-03
fix(aot): Handle the use case when the video is not available

hristoterezov:
Depends on https://github.com/jitsi/jitsi-meet/pull/2242


alwaysontop/render.js
```diff
@@ -52,11 +69,7 @@ class AlwaysOnTop {
      * @returns {HTMLElement|undefined} the large video.
      */
     get _jitsiMeetLargeVideo() {
-        const iframe = this._api.getIFrame();
-        if(!iframe || !iframe.contentWindow || !iframe.contentWindow.document) {
-            return;
-        }
-        return iframe.contentWindow.document.getElementById('largeVideo');
+        return this._api._getLargeVideo();
```
virtuacoplenny: Does this work?

PR #16, opened 2017-12-03
fix(aot): Improves the logic around transfering the Browser Window id

hristoterezov:


alwaysontop/main.js
```diff
@@ -53,9 +53,9 @@ function onAlwaysOnTopWindow(
         });
         jitsiMeetWindow.webContents.send('jitsi-always-on-top', {
             type: 'event',
-            name: 'new-window',
             data: {
-                id: win.id
+                id: win.id,
+                name: 'new-window'
```
virtuacoplenny: What is the advantage of putting name within data?

hristoterezov: Just to be consistent with Jitsi Meet Transport format.

PR #17, opened 2017-12-08
feat(aot): Restore the last known position of AOT window.

hristoterezov:


alwaysontop/main.js
```diff
@@ -62,13 +70,56 @@ function onAlwaysOnTopWindow(
...
+        screenRectangle) {
+    const maxY
+        = screenRectangle.y + screenRectangle.height - windowRectangle.height;
+    const maxX
+        = screenRectangle.x + screenRectangle.width - windowRectangle.width;
+
+    return {
+        x: Math.min(Math.max(windowRectangle.x, screenRectangle.x), maxX),
+        y: Math.min(Math.max(windowRectangle.y, screenRectangle.y), maxY)
+    };
```
virtuacoplenny: I think logic could use some comments because I'm having a hard time understanding with confidence what is happening. Could you also specify if a screens position origin is bottom-left/top-right/wherever?

alwaysontop/main.js
```diff
@@ -62,13 +70,56 @@ function onAlwaysOnTopWindow(
...
+}
+
+/**
+ * Returns the stored coordinates for the always on top window of its previous
+ * location or if the last location is unknown returns coordinates to display
+ * the window in the top right corner of the screen.
```
virtuacoplenny: When I tested I was seeing it in the top left. Is there an issue with the logic or the comment?

hristoterezov: it should be the logic!

PR #18, opened 2017-12-14
fix(build): do not whitelist object literal spread in eslint

virtuacoplenny:
It is not a standard feature right now and cannot be used
in stock node.


virtuacoplenny:
Nevermind, this is gonna get lumped into wifistats work.

PR #20, opened 2017-12-18
fix(aot): Remove listeners if there is an error with opening the AOT window

hristoterezov:


alwaysontop/render.js
```diff
@@ -208,6 +208,14 @@ class AlwaysOnTop {
...
         this._alwaysOnTopWindow = window.open(alwaysOnTopURL, 'AlwaysOnTop');
         if(!this._alwaysOnTopWindow) {
+            ipcRenderer.removeListener(
+                'jitsi-always-on-top',
+                this._onMessageReceived
+            );
+            this._api.removeListener(
+                'largeVideoChanged',
+                this._updateLargeVideoSrc
+            );
```
virtuacoplenny: Would it be simple to move all or most listener cleanup into a separate function and call that here instead and when closing the always on top window? I think it's preferable to have listener and cleanup steps grouped together as much as possible to better keep track of them.

virtuacoplenny:
Does this PR need to be updated?

PR #23, opened 2018-06-06
feat(google-auth): expose methods to enable opening google auth

virtuacoplenny:
- Expose a setup method to allow the jitsi iframe to open a new
  window, which listens for a specific redirect URL and passes
  back the full redirect URL.
- Expose a setup method that puts a global on the jitsi iframe
  to exercise opening the new window.
- Expose teardown methods to ensure listeners get cleaned up.

Related jitsi-meet changes: https://github.com/jitsi/jitsi-meet/pull/3106
Related jitsi-meet-electron changes: https://github.com/jitsi/jitsi-meet-electron/pull/39

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,27 +13,35 @@ The Google API utility exposes methods for performing Google client-side authent
 
 **Enable the Google API integration:**
 
-In the **main** electron process:
+To set up the Google API integration, in the **main** electron process, execute `setupGoogleApiMain`. This will set up listeners on the main processes to be aware of requests to open a Google popup.
 ```Javascript
 const { setupGoogleApiMain } = require("jitsi-meet-electron-utils");
 const options = {
     browserWindowOptions: {
-        webPreferences: {
-            nodeIntegration: false
-        }
+        ...
     }
 };
 setupGoogleApiMain(options);
 ```
 
-In the **render** electron process of the window where Jitsi Meet is displayed:
+The function `setupGoogleApiMain` takes in an options objects the supports the following:
+- `browserWindowOptions`: The BrowserWindow options object to pass into the Google authentication popup. These are the same options as supported by the [BrowserWindow class](https://github.com/electron/electron/blob/master/docs/api/browser-window.md#class-browserwindow). By default the following options will be used:
+    ```
+    {
+        webPreferences: {
+            nodeIntegration: false
+        }
+    }
+    ```
+
+To finish setting up the Google API integration, in the **render** electron process of the window where Jitsi Meet is displayed, execute `setupGoogleApiRender` to expose globals to Jitsi-Meet to interact indirectly with the main processes.
 ```Javascript
 const { setupGoogleApiRender } = require("jitsi-meet-electron-utils");
 const api = new JitsiMeetExternalAPI(...);
 setupGoogleApiRender(api);
 ```
 
-To clean up Google API integration, in the **main** electron process:
+To clean up various listeners attached by the Google API integration, in the **main** electron process:
 ```
 teardownGoogleApi();
 ```
--- a/README.md
+++ b/README.md
@@ -7,6 +7,35 @@ jitsi-meet-electron-utils contains native code for some utilities. You'll need [
 NOTE: For Linux install libxtst-dev and libpng++-dev (`sudo apt-get install libxtst-dev libpng++-dev`). This dependancies are related to RobotJS which is a dependency of jitsi-meet-electron-utils. You can see the build instructions for RobotJS [here](https://github.com/jitsi/robotjs/tree/jitsi#building)
 
 ## Usage
+#### Google API
+The Google API utility exposes methods for performing Google client-side authentication by opening a oauth popup window and passing back the redirect URL with the Google access token.
+
+**Enable the Google API integration:**
+
+In the **main** electron process:
+```Javascript
+const { setupGoogleApiMain } = require("jitsi-meet-electron-utils");
+const options = {
+    browserWindowOptions: {
+        webPreferences: {
+            nodeIntegration: false
+        }
+    }
+};
+setupGoogleApiMain(options);
+```
+
+In the **render** electron process of the window where Jitsi Meet is displayed:
+```Javascript
+const { setupGoogleApiRender } = require("jitsi-meet-electron-utils");
+const api = new JitsiMeetExternalAPI(...);
+setupGoogleApiRender(api);
+```
+
+To clean up Google API integration:
+```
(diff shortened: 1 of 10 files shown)


README.md
```diff
@@ -7,6 +7,35 @@ jitsi-meet-electron-utils contains native code for some utilities. You'll need [
...
+In the **render** electron process of the window where Jitsi Meet is displayed:
+```Javascript
+const { setupGoogleApiRender } = require("jitsi-meet-electron-utils");
+const api = new JitsiMeetExternalAPI(...);
+setupGoogleApiRender(api);
+```
+
+To clean up Google API integration:
+```
+teardownGoogleApi();
```
hristoterezov: This line should be called in the main process or the render process or it doesn't matter?

googleapi/constants.js
```diff
@@ -0,0 +1,4 @@
+module.exports = {
+    AUTH_FINISHED: 'jitsi-google-auth-finish',
```
hristoterezov: JS docs missing

README.md
```diff
@@ -7,6 +7,35 @@ jitsi-meet-electron-utils contains native code for some utilities. You'll need [
...
+#### Google API
+The Google API utility exposes methods for performing Google client-side authentication by opening a oauth popup window and passing back the redirect URL with the Google access token.
+
+**Enable the Google API integration:**
+
+In the **main** electron process:
+```Javascript
+const { setupGoogleApiMain } = require("jitsi-meet-electron-utils");
+const options = {
+    browserWindowOptions: {
```
hristoterezov: Aren't this the options by default? What's the point of passing them?

googleapi/main.js
```diff
@@ -0,0 +1,112 @@
...
+
+/**
+ * Removes listeners on the main Electron process for opening new windows for
+ * getting a Google access token.
+ *
+ * @returns {void}
+ */
+function teardownGoogleApiMain() {
+    ipcMain.removeAllListeners(constants.AUTH_REQUESTED);
+    browserWindowOptions = defaultBrowserWindowOptions;
```
hristoterezov: Do we need this line since on `setupGoogleApiMain` method `browserWindowOptions ` value would be reseted anyway?

googleapi/main.js
```diff
@@ -0,0 +1,112 @@
...
+ * @param {Object} options - Additional configuration for setting up the
+ * listeners.
+ * @param {Object} browserWindowOptions - The options to pass into the
+ * constructor for {@code BrowserWindow}.
+ * @returns {void}
+ */
+function setupGoogleApiMain(options = {}) {
+    teardownGoogleApiMain();
+
+    browserWindowOptions
```
hristoterezov: Actually why do we need to expose the BrowserWindow options? Do you have an use case in mind where the consumer of the library will need to change anything there?

virtuacoplenny: I don't know what the consumer needs. For example, the consumer might have security restrictions on what options needs to be set on new windows. So to reduce the need to make changes in utils I just opened up options.

googleapi/render.js
```diff
@@ -0,0 +1,100 @@
...
+    iframe && iframe.addEventListener('load', setGlobalGoogleApiVariable);
+}
+
+/**
+ * Removes listeners that may have been set for interacting with the Google API.
+ *
+ * @returns {void}
+ */
+function teardownGoogleApiRender() {
+    if (iframe) {
```
hristoterezov: What do you think about the idea to replace the attaching of the method to the iframe window's DOM with a "private" listener on the iframe API?

IMHO using the iframe API vs directly manipulating the DOM looks cleaner to me. Also in this case the whole logic in the render process is to pass and receive IPC messages and real implementation is in the main process.

virtuacoplenny: Okay, so take in the api and add a listener through some kind of api.addListener method? Does that method exist already? I'd want to change the screenshare setup as well then because I re-used its approach.

googleapi/main.js
hristoterezov: I'm with mixed feelings about this but this make sense.

README.md
virtuacoplenny: Hmm...Yeah, should be called in the main process because it removes listeners that are applied to ipc main. I'll specify that

README.md
virtuacoplenny: Those are the default options. I just wanted to pass an example of passing browserWindowOptions. Do you think that's unclear for a doc? Should I use pseudo code instead to make clear that electron browser window options can be passed?

googleapi/main.js
virtuacoplenny: I suppose you don't need it if you assume setupGoogleApiMain will be called. But I also wanted to just be balanced with how I set browserWindowOptions to the defaults at the top. It's probably best to just null it out in both cases.

googleapi/main.js
virtuacoplenny: I'm open to changes here. Request them publicly or privately. I can make some assumptions of what will be used by the consumer I'm building this out for specifically.

googleapi/render.js
hristoterezov: So you can probably use `on`. Although probably new method specifically for this use case would be better because you need a request from meet. You can use the `transport` request mechanism maybe. 

I'm not sure if it worth the effort! This is only my personal preference, so if you like it better than the current approach, let's do it, otherwise we can keep the current logic.

For the screen sharing I'd like to see this change! Unfortunately I'm not sure if it is not too late because of backward compatibility issues. I think this is the reason for using `window.JitsiMeetElectron` vs `window.JitsiMeetJS` which is our namespace for globals lately.

googleapi/main.js
hristoterezov: But without calling `setupGoogleApiMain` it should fail right? I like the idea about `null`. Isn't better to use `undefined` instead of `null` in all those cases?

googleapi/main.js
hristoterezov: I don't have better idea currently. Let's leave like it is!

googleapi/main.js
virtuacoplenny: Sure. It's about the same to me because it's a private variable and a coerced truthy check is done. null is probably better in the case when passing around a falsy value to signify it was not an intentional undefined. So I can just use undefined here.

googleapi/render.js
hristoterezov: Actually please hold on on this one because I'm reviewing the meet part right now and I may discard the comment :)

virtuacoplenny:
TODO: increase version number in package.json

PR #28, opened 2018-07-24
Fix error with mouse moving in Windows with multiple displays

vyatkin0:
There is a error in robotjs that causes incorrect mouse position in Windows 10 with multiple displays.
RobotJS doesn't  work when displays have different dpi or placed on virtual screen randomly.

### DIFF
--- a/node_addons/sourceId2Coordinates/index.js
+++ b/node_addons/sourceId2Coordinates/index.js
@@ -1,6 +1,4 @@
-const sourceId2Coordinates
-    = require("../../build/Release/sourceId2Coordinates.node")
-        .sourceId2Coordinates;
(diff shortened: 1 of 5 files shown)


saghul:
Thanks Aleksey! @hristoterezov can you PTAL?

vyatkin0:
The problem with screen sharing is still exists. Jitsi reports wrong display number to the remote control in Windows. For example, it shares display 0 but reports to remote control that display 1 is shared. The problem is related to electron->chromium->webrtc framework and solved in electron 3.0 beta2. With the latest version with electron Jitsi works well.

hristoterezov:
@vyatkin0 Have you tested with the latest versions jitsi-meet-electron-utils?

saghul:
No follow-up, closing.

PR #29, opened 2019-01-31
feat(alwaysontop): Add visibility monitoring for meeting frame

gimre:


alwaysontop/render.js
```diff
@@ -214,6 +244,7 @@ class AlwaysOnTop {
             api: this._api,
             onload: this._updateLargeVideoSrc,
             onbeforeunload: () => {
+                this._api.emit(ALWAYSONTOP_WILL_CLOSE);
```
hristoterezov: Currently we don't have any events or public interfaces in the IFrame API related to the always on top. Also I think this will fit better to the jitsi-meet-electron-utils API.

Could you please not emit this trough the IFrame API instance but handle it somehow in the jitsi-meet-electron-utils project? 
One option that I can think of is to pass an "always on top will close" callback to `setupAlwaysOnTopRender`. Another one will be to return an `Emitter` from the `setupAlwaysOnTopRender` which can be used to emit the `ALWAYSONTOP_WILL_CLOSE` event.  Maybe there are even better options...

Also could you please add some information about this in the README.md?

gimre: Changed the initial piggybacking on the external api to an EventEmitter instance. Added some extra fixes for corner cases aswell.

alwaysontop/render.js
```diff
@@ -244,20 +277,28 @@ class AlwaysOnTop {
...
+                    x: position[0],
+                    y: position[1]
+                };
+            }
         }
+
         if(this._alwaysOnTopWindow) {
-            this._alwaysOnTopWindow.close();
-            this._alwaysOnTopWindow = undefined;
+            if(!this._alwaysOnTopBrowserWindow ||!this._alwaysOnTopBrowserWindow.isDestroyed()) {
```
hristoterezov: Do we need this checks here? Why?

alwaysontop/render.js
```diff
@@ -244,20 +277,28 @@ class AlwaysOnTop {
...
     _closeAlwaysOnTopWindow() {
         if (this._alwaysOnTopBrowserWindow) {
-            const position
-                = this._alwaysOnTopBrowserWindow.getPosition();
-            this._alwaysOnTopBrowserWindow = undefined;
-            this._position = {
-                x: position[0],
-                y: position[1]
-            };
+            if (!this._alwaysOnTopBrowserWindow.isDestroyed()) {
```
hristoterezov: Why not add this condition to the upper `if`?

Just out of curiosity why did you add this check?

alwaysontop/render.js
gimre: there is a problem with timing in electron if you try to close a window multiple times in a row. if you issue two calls to `.close()` on the javascript window object, the second call will throw a RPC exception. Also, the window.closed property is not set to true immediately after the first call so there is no reliable way of preventing the exception. Checking BrowserWindow.isDestroyed() is robust though.

gimre: oh, I misread the question... experimentally, yes, I guess there is some problem with how those references are being handles, but that is a separate, pre-existing issue.

alwaysontop/render.js
gimre: there is a problem with timing in electron if you try to close a window multiple times in a row. if you issue two calls to `.close()` on the javascript window object, the second call will throw a RPC exception. Also, the window.closed property is not set to true immediately after the first call so there is no reliable way of preventing the exception. Checking BrowserWindow.isDestroyed() is robust though.

gimre: i only wanted to guard the close call to prevent tampering with existing functionality.

alwaysontop/render.js
hristoterezov: Experimentally? How? Could you please elaborate?

hristoterezov:
@gimre I got a little confused here. I think your answers somehow got mixed up between the two questions! Are they?

Probably we'll need some comments in the code....

alwaysontop/render.js
gimre: updated :D

PR #30, opened 2019-02-14
feat(wifistats): Change to sync write in iframe window object

gimre:


damencho:
Does this change affects jitsi-meet-electron? I mean have you tested it there, maybe after this change we need to add the load listener there, if there are cases which brakes it?

gimre:
@damencho it shouldn't affect anything, but I wrote to @hristoterezov and hopefully he can test it in jitsi-meet-electron to confirm

damencho:
One more question, why this listener works for always on top, remote control, and screen sharing but will not work with wifi stats?

hristoterezov:
I guess the listener handles reloads of the iframe due to non-recoverable errors. IMHO the listener have to be there and in addition we should execute it or add into the docs that the function needs to be called before the iframe is loaded.

gimre:
talked to @hristoterezov and he will be implementing a more robust solution. closing for now

PR #39, opened 2019-04-23
fix: Alwaysontop window crash

gimre:


alwaysontop/helpers.js
```diff
@@ -0,0 +1,36 @@
...
+  let tries = 5;
+  let win;
+
+  while (!isValidWindow(win) && tries > 0) {
+    tries -= 1;
+
+    win = window.open(...args);
+
+    // eslint-disable-next-line
+    await waitEventRounds(2);
```
hristoterezov: Why do we need the retry logic? Have you noticed failures to open window?

alwaysontop/helpers.js
```diff
@@ -0,0 +1,36 @@
+// eslint-disable-file
```
hristoterezov: Why eslint is disabled?

alwaysontop/helpers.js
gimre: yes, electron will open a window sometimes, but it will be in limbo and there will be an IPC error so the loop circumvents this

alwaysontop/helpers.js
gimre: Forgot to remove that - updated

gimre:
Added jsdoc.

alwaysontop/helpers.js
```diff
@@ -0,0 +1,59 @@
...
+const waitOneEventRound = () => new Promise(y => setTimeout(y));
+
+/**
+ * Returns a promise that resolves after *count* event loop passes.
+ *
+ * @param {number} count - How many event loop passes to wait
+ * @returns {Promise<void>}
+ */
+const waitEventRounds = async count => {
+  // eslint-disable-next-line
```
hristoterezov: Why disable eslint?

alwaysontop/helpers.js
```diff
@@ -0,0 +1,59 @@
...
+  for (let i = count; i > 0; i --) {
+    // eslint-disable-next-line
```
hristoterezov: Why disable eslint?

alwaysontop/helpers.js
hristoterezov: And why do we need the timeout?

alwaysontop/helpers.js
```diff
@@ -0,0 +1,59 @@
+/**
```
hristoterezov: Could you please rename the file to functions.js? This way it will be more consistent with jitsi-meet file names.

alwaysontop/helpers.js
```diff
@@ -0,0 +1,59 @@
...
+ * Validates a window opened via window.open
+ *
+ * @returns {boolean} - Validity of the window
+ */
+const isValidWindow = win => win && !win.closed;
```
hristoterezov: Could you please define the functions as `function ...` and not as `const ...`? This way it will be more consistent with the code from jitsi-meet.

alwaysontop/render.js
```diff
@@ -239,36 +230,56 @@ class AlwaysOnTop extends EventEmitter {
...
-             * (-webkit-app-region: drag) all mouse events are blocked. To fix
-             * this we'll implement drag ourselves.
-             */
-            shouldImplementDrag: os.type() !== 'Darwin',
-            move: (x, y) => {
-                if (this._alwaysOnTopBrowserWindow) {
-                    this._alwaysOnTopBrowserWindow.setPosition(x, y);
+
+        openWindow('', 'AlwaysOnTop')
+            .then(win => {
```
hristoterezov: Shouldn't we add a `catch`?

hristoterezov:
@gimre When I tested it with jitsi-meet-electron half of the times when the always on top window is displayed stays white. It seems there isn't any HTML injected.

alwaysontop/render.js
gimre: this promise is never rejected

hristoterezov:
Fixed with 329f249f33d03eee39fdbf1b9ce811588ed4a614

PR #40, opened 2019-05-01
Fix remote control mouse coordinates issue

hristoterezov:


remotecontrol/index.js
```diff
@@ -45,13 +46,25 @@ class RemoteControl {
     }
 
     /**
-     * Handles remote control start messages.
+     * Returns the scale factor for the current display used to calculate the resolution of the display.
+     *
+     * NOTE: On Mac OS this._display.scaleFactor will always be 2 for some reason. But the values returned from
```
saghul: are you sure this holds true for non-retina macs, external displays or macs with the display configured for "more space" ?

hristoterezov: I tested with external non-retina display and it was fine! I tested both over-scaling  and under-scaling and it was working!

I'm not very worried about it because this actually preserves the old Mac OS behaviour. So even if we missed use case we won't introduce regression and if we find exception we can add additional conditions.

saghul: 👍

PR #41, opened 2019-05-02
fix(remote-control): Position of the display.

hristoterezov:


remotecontrol/index.js
```diff
@@ -182,8 +200,8 @@ class RemoteControl {
             case EVENTS.mousemove: {
                 const { width, height, x, y } = this._display.bounds;
                 const scaleFactor = this._getDisplayScaleFactor();
-                const destX = (data.x * width + x) * scaleFactor;
-                const destY = (data.y * height + y) * scaleFactor;
+                const destX = data.x * width * scaleFactor + x;
+                const destY = data.y * height * scaleFactor + y;
```
virtuacoplenny: I'm math illiterate. Can you tell me what this fixes?

hristoterezov: Sure! 

`data.x` and `data.y` are received from the controller and they represent the relative position of the mouse to the video of the desktop sharing stream divided by the size of the video. So in order to calculate the position of the mouse on original screen we need to generally multiply by the width of the screen. The actual width of the screen will be the `width` that we get from electron multiplied by the scale factor of the display (if any).

The `x` and `y` are going to be different than `0` only if there are multiple monitors. In the multiple monitor situation the way we generally set the mouse coordinates is the following:
 - There is a desktop area which is big rectangle that includes all the displays there.
 - The coordinates that we are setting are related to the whole desktop area.
`data.x * width * scaleFactor` will be the coordinates related to the specific display. In order to convert those coordinates to the 'desktop area' coordinates we need to add the `x`/`y`.

PR #42, opened 2019-05-02
fix(always-on-top): use about:blank for always on top

virtuacoplenny:
The always on top window can lose reference to the
opener and any global variables set on it. To
get around this, do not navigate away from the
initial about:blank page that loads when the window
opens and instead use it as the always on top window.


virtuacoplenny:
> I'm wondering if it makes sense to add a comment in the code why do we use about:blank and what are the issues around the file URL. WDYT?

Are the comment I already added not enough or not in the right place?

alwaysontop/render.js
```diff
@@ -269,6 +263,31 @@ class AlwaysOnTop extends EventEmitter {
                 }
             }
         };
+
+        // Change the dom of 'about:blank' to display the always on top content.
+        this._alwaysOnTopWindow.onload = () => {
```
gimre: this seems to do the trick! kudos!

gimre: i'm so happy, i'm almost crying

PR #45, opened 2019-07-02
Make always on top window resizable

horymury:
- minimum size is the starting size of the window
- aspect ratio of video gets preserved on resize, and empty spaces either on top-bottom/left-right are gray background. this is discussed and agreed with PM

### DIFF
--- a/alwaysontop/main.js
+++ b/alwaysontop/main.js
@@ -1,7 +1,14 @@
 const os = require('os');
 const electron = require('electron');
+const robot = require("robotjs");
 const { BrowserWindow, ipcMain } = electron;
 const { SIZE } = require('./constants');
+
+/**
+ * The aspect ratio to preserve during AOT window resize
+ * @type {number}
+ */
+const ASPECT_RATIO = 16 / 9;
 
 /**
  * The coordinates(x and y) of the always on top window.
@@ -9,6 +16,17 @@ const { SIZE } = require('./constants');
  * @type {{x: number, y: number}}
  */
 let position = {};
+
+/**
+ * Stores the current size of the AOT during the conference
+ * @type {{width: number, height: number}}
+ */
+let size = Object.assign({}, SIZE);
+
+/**
+ * Keeps the old size of the window between resize handler calls
+ */
+let oldSize;
 
 /**
  * Handles new-window events for the main process in order to customize the
@@ -34,15 +52,11 @@ function onAlwaysOnTopWindow(
         const win = event.newGuest = new BrowserWindow(
             Object.assign(options, {
                 backgroundColor: 'transparent',
-                width: SIZE.width,
-                height: SIZE.height,
                 minWidth: SIZE.width,
                 minHeight: SIZE.height,
-                maxWidth: SIZE.width,
-                maxHeight: SIZE.height,
                 minimizable: false,
                 maximizable: false,
-                resizable: false,
+                resizable: true,
                 alwaysOnTop: true,
                 fullscreen: false,
                 fullscreenable: false,
@@ -50,13 +64,44 @@ function onAlwaysOnTopWindow(
                 titleBarStyle: undefined,
                 frame: false,
                 show: false
-            }, getPosition())
+            }, getPosition(), getSize())
         );
         win.once('ready-to-show', () => {
             if (win && !win.isDestroyed()) {
                 win.showInactive();
             }
         });
+
+        //for macOS we use the built-in setAspectRatio on resize, for other we use custom implementation
+        if (os.type() === 'Darwin') {
+            win.setAspectRatio(ASPECT_RATIO);
+        } else {
+            win.on('will-resize', (e, newBounds) => {
+                oldSize = win.getSize();
+                const mousePos = robot.getMousePos();
+                const windowBottomRightPos = {
+                    x: newBounds.x + newBounds.width - 16,
+                    y: newBounds.y + newBounds.height - 16,
+                };
+                //prevent resize from bottom right corner as it is buggy.
+                if (mousePos.x >= windowBottomRightPos.x && mousePos.y >= windowBottomRightPos.y) {
+                    e.preventDefault();
+                }
+            });
+            win.on('resize', () => {
+                let [width, height] = win.getSize();
+                //we scale either width or height according to the other by checking which of the 2
+                //changed the most since last resize.
+                if (Math.abs(oldSize[0] - width) >= Math.abs(oldSize[1] - height)) {
+                    height = Math.round(width / ASPECT_RATIO);
+                } else {
+                    width = Math.round(height * ASPECT_RATIO);
+                }
+                win.setSize(width, height);
+                size.width = width;
+                size.height = height;
+            });
+        }
         jitsiMeetWindow.webContents.send('jitsi-always-on-top', {
             type: 'event',
             data: {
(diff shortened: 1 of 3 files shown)


hristoterezov:
@horymury I noticed the following behaviour:

1. Every time when the always on top window is reopened (even during the same call) the size is reset to the default one.
2. The window can be resized to a different aspect ratio. We don't have any limitations for the aspect ratio.

I just want to make sure that this is not a problem.

Otherwise looks good to me and if you confirm that the above is the desired behaviour I'll merge the PR.

horymury:
@hristoterezov Emil had the following requests:

1. see if we can find out why AOT window keeps growing on moving it - Windows only
2. keep the last set size on the same meeting session or even the same VOD session
3. keep the aspect ratio on resize

alwaysontop/render.js
```diff
@@ -234,19 +263,38 @@ class AlwaysOnTop extends EventEmitter {
...
+                previousHeight = size[1];
+            },
             onbeforeunload: () => {
                 this.emit(ALWAYSONTOP_WILL_CLOSE);
                 this._api.removeListener(
                     'largeVideoChanged',
                     this._updateLargeVideoSrc
                 );
             },
+            onresize: debounce(() => {
```
hristoterezov: Maybe for Mac you can use https://electronjs.org/docs/api/browser-window#winsetaspectratioaspectratio-extrasize-macos

hristoterezov: Why do we need the debounce?

alwaysontop/render.js
```diff
@@ -304,12 +352,17 @@ class AlwaysOnTop extends EventEmitter {
                 x: position[0],
                 y: position[1]
             };
+
+            const size = this._alwaysOnTopBrowserWindow.getSize();
+            const width = size[0];
+            const height = size[1];
+            this._sendSize({ width, height });
```
hristoterezov:
```
this._sendSize({
    width: size[0],
    height: size[1]
})
```

Isn't this looking better?

alwaysontop/render.js
```diff
@@ -146,6 +150,30 @@ class AlwaysOnTop extends EventEmitter {
         }
     }
 
+    _sendResetSize() {
```
hristoterezov: Missing JSDoc.

alwaysontop/render.js
```diff
@@ -146,6 +150,30 @@ class AlwaysOnTop extends EventEmitter {
...
+        ipcRenderer.send('jitsi-always-on-top', {
+            type: 'event',
+            data: {
+                name: 'size',
+                type: 'reset',
+            }
+        });
+    }
+
+    _sendSize({ width, height }) {
```
hristoterezov: Missing JSDoc.

alwaysontop/render.js
```diff
@@ -1,12 +1,16 @@
...
 const { ipcRenderer, remote } = require('electron');
+const debounce = require('debounce');
 
 const { EventEmitter } = require('events');
 const os = require('os');
 const path = require('path');
 
 const { ALWAYSONTOP_WILL_CLOSE } = require('./constants');
 
+const aspectRatio = 1.7777777777777777;
```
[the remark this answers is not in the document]
hristoterezov: Also please add JSDoc.

alwaysontop/render.js
```diff
@@ -1,12 +1,16 @@
...
+let previousWidth;
```
hristoterezov: Missing JSDoc.

alwaysontop/render.js
```diff
@@ -1,12 +1,16 @@
...
+let previousHeight;
```
hristoterezov: Missing JSDoc.

alwaysontop/render.js
```diff
@@ -1,12 +1,16 @@
...
 /**
```
hristoterezov: Add new line

horymury:
> Left some comments and questions!
> 
> Also when I tested it the resize feels a little jumpy and choppy. I think there is a problem with the calculation logic. Could you please take a look?

The choppines and jumpiness where resolved by moving the logic to main process.

alwaysontop/render.js
horymury: removed debounce

horymury: I'm working on windows so it's a bit tricky to test it out. But with the changes I made it should work just fine with the current logic on MAC as well

alwaysontop/main.js
```diff
@@ -38,27 +56,55 @@ function onAlwaysOnTopWindow(
...
                 show: false
-            }, getPosition())
+            }, getPosition(), getSize())
         );
         win.once('ready-to-show', () => {
             win.showInactive();
         });
+
+        //for macOS we use the built-in setAspectRatio on resize, for other we use custom implementation
+        if (os.type() === 'Darwin') {
```
hristoterezov: Why did we move the resize code from the render to the main process?

alwaysontop/main.js
```diff
@@ -38,27 +56,55 @@ function onAlwaysOnTopWindow(
...
+            win.setAspectRatio(16/9);
+        } else {
+            win.on('will-resize', (e, newBounds) => {
+                oldSize = win.getSize();
+                const mousePos = robot.getMousePos();
```
hristoterezov: Do we really need to use robotjs for this? Isn't it https://github.com/electron/electron/blob/master/docs/api/screen.md#screengetcursorscreenpoint do the job?

alwaysontop/main.js
```diff
@@ -38,27 +56,55 @@ function onAlwaysOnTopWindow(
...
+                const windowBottomRightPos = {
+                    x: newBounds.x + newBounds.width - 16,
```
hristoterezov: Why `-16`?

alwaysontop/main.js
```diff
@@ -38,27 +56,55 @@ function onAlwaysOnTopWindow(
...
+                    y: newBounds.y + newBounds.height - 16,
+                };
+                //prevent resize from bottom right corner as it is buggy.
+                if (mousePos.x >= windowBottomRightPos.x && mousePos.y >= windowBottomRightPos.y) {
```
hristoterezov: I'm wondering isn't it there a better way to detect this? 

I'm worried that you can end up in the same situation even if you started the resize from another corner. Although probably we shouldn't care about this.

alwaysontop/main.js
```diff
@@ -1,19 +1,37 @@
...
 const SIZE = {
     width: 320,
     height: 180
 };
 
+/**
+ * The aspect ratio to preserve during AOT window resize
+ * @type {number}
+ */
+const aspectRatio = 16 / 9;
```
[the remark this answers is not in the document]
horymury: done

alwaysontop/main.js
horymury: as it seemed to be more fluent on main process

alwaysontop/main.js
horymury: I've set this as the diagonal resize does not necessarily start precisely on the window's right bottom corner, it has some pixels threshold on right and bottom window edges in which is still a diagonal resize

alwaysontop/main.js
horymury: Please see following image
First is robotjs detection, second is electron screen detection on a monitor having scale 125%. 
Please note that robotjs correctly shows the y axe as 0 in this case, and electron screen detects it wrongly as 200 something. I think this wrong detection electron-side causes all the bugs reported on Electron related to scaled monitors on Windows.

image

alwaysontop/main.js
horymury: Not sure if it s possible to end up in this situation from another corner, as the check is done on future position of the window,  and the future position doesn't seem possible to have the bottom right corner 'upper' and 'lefter' from current mouse position :)

PR #47, opened 2019-07-05
Updates setupWiFiStats to not depend on iframe load event.

damencho:


damencho:
We decided offline that will add a change in jitsi-meet with a wrapper function of getWiFiStats

PR #48, opened 2019-07-09
Adds sending power monitor events to the jitsi-meet iframe.

damencho:


powermonitor/render.js
```diff
@@ -0,0 +1,105 @@
...
+    }
+}
+
+/**
+ * Initializes the power monitor in the render process of the
+ * window which displays Jitsi Meet.
+ *
+ * @param iframe - the iframe to send events to.
+ */
+module.exports = function setupPowerMonitorRender(iframe) {
```
hristoterezov: I know we require the iframe directly as a parameter for some of our electron APIs but isn't it going to be better to require the api instance and get iframe from it. Looks a little bit more easier for the consumers and also if we ever need something from the api in future, we won't need to change the electron API (this happened to me once in the past).

powermonitor/render.js
```diff
@@ -0,0 +1,105 @@
...
+        _channel = postis({
+            window: iframe.contentWindow,
+            windowForEventListening: window,
+            scope: 'jitsi-power-monitor'
+        });
+        _channel.ready(() => {
+            _channel.listen('message', message => {
+                const { name } = message.data;
+                if(name === POWER_MONITOR_MESSAGE_NAME) {
+                    ipcRenderer.send(POWER_MONITOR_QUERIES_CHANNEL, message);
```
hristoterezov: We are sending a message to us here?

powermonitor/render.js
```diff
@@ -0,0 +1,105 @@
...
+ * The channel we use to communicate with Jitsi Meet window.
+ */
+let _channel;
+
+/**
+ * The listener for query responses.
+ * @param _ - Not used.
+ * @param response - The response to send.
+ */
+let queriesChannelListener = function (_, response) {
```
hristoterezov: Why this function definition and the next one are different than all other function definitions in the PR?

powermonitor/render.js
damencho: Ok, will change it.

powermonitor/render.js
damencho: Yep, this sends a message back to the main process.

powermonitor/render.js
damencho: Cause these a the listener we want to remove later.

powermonitor/render.js
```diff
@@ -0,0 +1,105 @@
...
+    _sendEvent(event);
+};
+
+/**
+ * Sends event to Jitsi Meet.
+ *
+ * @param {Object} event the remote control event.
+ */
+function _sendEvent(event) {
+    const powerMonitorEvent = Object.assign(
```
hristoterezov: Redundant variable.

powermonitor/main.js
```diff
@@ -0,0 +1,91 @@
...
+}
+
+/**
+ * Initializes the power monitor functionality in the main electron process.
+ *
+ * @param {BrowserWindow} jitsiMeetWindow - the BrowserWindow object which
+ * displays Jitsi Meet
+ */
+module.exports = function setupPowerMonitorMain(jitsiMeetWindow) {
+    app.isReady() ? _attachEvents(jitsiMeetWindow)
```
hristoterezov: Optional: I'd use the good old `if` here.

powermonitor/render.js
```diff
@@ -0,0 +1,105 @@
...
+        method: 'message',
+        params: message
+    });
+}
+
+/**
+ * Disposes the power monitor functionality.
+ */
+function dispose() {
+
```
hristoterezov: Redundant empty line

powermonitor/render.js
```diff
@@ -0,0 +1,105 @@
...
+            'unload',
+            dispose
+        );
+        _channel = postis({
+            window: iframe.contentWindow,
+            windowForEventListening: window,
+            scope: 'jitsi-power-monitor'
+        });
+        _channel.ready(() => {
+            _channel.listen('message', message => {
```
hristoterezov: Do we receive such a messages?

damencho: yes

PR #50, opened 2019-07-16
Fix window resize on move

horymury:
This fixes :
https://github.com/electron/electron/issues/9477

On monitors having scale options applied,window.setPosition() also cause a growth of the window. 
The fix consists in resetting to initial window size once the move/drag event finished.
This only applies on Windows.


hristoterezov:
Have you tried to use setBounds as described in https://github.com/electron/electron/issues/9477 ?

horymury:
@hristoterezov setBounds worked it now works flawlessly on scaled monitor. thanks

PR #56, opened 2019-09-26
fix(electron5): support

hristoterezov:


alwaysontop/render.js
```diff
@@ -98,7 +98,7 @@ class AlwaysOnTop extends EventEmitter {
      */
     get _alwaysOnTopWindowVideo() {
         if (!this._alwaysOnTopWindow || !this._alwaysOnTopWindow.document) {
-            return;
+            return undefined;
```
saghul: This looks unnecessary.

alwaysontop/render.js
```diff
@@ -304,19 +312,13 @@ class AlwaysOnTop extends EventEmitter {
...
-            }
-
             const cssPath = path.join(__dirname, './alwaysontop.css');
             const jsPath = path.join(__dirname, './alwaysontop.js');
 
             // Add the markup for the JS to manipulate and load the CSS.
             this._alwaysOnTopWindow.document.body.innerHTML = `
               <div id="react"></div>
-              <video autoplay="" id="video" style="transform: none;"></video>
+              <video autoplay="" id="video" style="transform: none;" muted="muted"></video>
```
saghul: I don't think a value is necessary for the `muted` property

popupsconfig/main.js
```diff
@@ -28,7 +28,9 @@ function initPopupsConfiguration(jitsiMeetWindow) {
             event.newGuest = new BrowserWindow(Object.assign(options, {
                 titleBarStyle: undefined,
                 webPreferences: {
-                    nodeIntegration: false
+                    nodeIntegration: false,
+                    contextIsolation: false,
```
saghul: not sure if we alphabeticallty sort in this project, but here goes nothing

alwaysontop/render.js
hristoterezov: ESLint complained....

alwaysontop/render.js
hristoterezov: Will try without it.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "jitsi-meet-electron-utils",
-  "version": "0.6.8",
+  "version": "0.6.9",
```
saghul: Make it 1.0 already!

hristoterezov: We should! But let's wait for the next minor bump at least?

PR #57, opened 2019-10-04
feat(electron6): support

hristoterezov:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "jitsi-meet-electron-utils",
-  "version": "0.6.9",
+  "version": "1.0.0",
```
saghul: It's happening!

powermonitor/main.js
```diff
@@ -65,19 +65,29 @@ module.exports = function setupPowerMonitorMain(jitsiMeetWindow) {
...
         switch(data.type) {
             case METHODS.queryIdleState:
-                electron.powerMonitor.querySystemIdleState(
-                    data.idleThreshold,
-                    idleState => {
-                        systemIdleResult(id, idleState);
-                    });
+                if (typeof powerMonitor.getSystemIdleState === 'function') { // electron 5+
+                    systemIdleResult(id, powerMonitor.getSystemIdleState(data.idleThreshold));
+                } else { // electron 4 or older
```
saghul: Question: do we need to support Electron < 5?

hristoterezov: We do unfortunately! We even have production use case for this!

We could probably just leave it unsupported right now and tell anybody who wants to use electron 4 to use the old release but I'm worried that we may need to push fixes for the old releases if we go down this road. This will be very hard for maintenance.... Supporting it for now looks easier...

Also upgrading to electron 5 with node 12 for projects with a lot of native add-ons turns out to be not so trivial task. So I guess it is good to support it for some time in order to give consumers more time to upgrade electron. Currently this is the only part of the code which needs adjustment.

PR #64, opened 2020-01-28
feat(alwaysontop): add dismiss button and event

gimre:


alwaysontop/main.js
```diff
@@ -64,9 +64,9 @@ function onAlwaysOnTopWindow(
                 titleBarStyle: undefined,
                 frame: false,
                 show: false,
-                webPreferences: {
+                webPreferences: Object.assign({}, options.webPreferences, {
```
hristoterezov: Why?

alwaysontop/constants.js
```diff
@@ -1,4 +1,5 @@
 module.exports = {
+    ALWAYSONTOP_DISMISSED: 'dismissed',
```
hristoterezov: Do we need this event? Why?

alwaysontop/render.js
```diff
@@ -319,9 +319,17 @@ class AlwaysOnTop extends EventEmitter {
             this._alwaysOnTopWindow.document.body.innerHTML = `
               <div id="react"></div>
               <video autoplay="" id="video" style="transform: none;" muted></video>
+              <div class="dismiss">&times;</div>
               <link rel="stylesheet" href="file://${ cssPath }">
             `;
 
+            this._alwaysOnTopWindow.document.body.addEventListener('click', e => {
```
hristoterezov: isn't it better to attach the listener on the element instead of the body?

alwaysontop/constants.js
gimre: to know when the window was dismissed

alwaysontop/main.js
gimre: to keep webPreferences set upstream, not overwrite them

alwaysontop/render.js
gimre: in this case this is more explicit, because we are setting innerHTML

alwaysontop/main.js
hristoterezov: I figured that. My question is more like:
How this relates to the PR? Are you trying yo solve an issue with this? What is the issue? What is the need of this change?

If this is not related to the PR, please open another PR with this change and put some description there what are you trying to solve!

alwaysontop/constants.js
hristoterezov: How are you consuming this event?

gimre: i am consuming it from the requiring module to have logic that would prevent the window being displayed if it is dimissed by the user

alwaysontop/main.js
gimre: it was just something I noticed while trying different approaches - I'll revert this change

alwaysontop/render.js
hristoterezov: ok but performance-wise, isn't it going to be better to put the lister on the element?

Also I'm wondering if alwaysontop.js is better place for this code. WDYT?

gimre: doesn't matter tbh but found a workaround

gimre:
> Left some comments and questions!
> 
> Also when I tested the 'x' is not aligned very well, it's closer to the bottom of the circle:
> Screenshot 2020-01-29 at 11 05 52 AM

what OS are you using? kind of looks like some font problem somehow

gimre:
seems `display: flex` and text alignment have some quirks when using different fonts - reimplemented without using text

alwaysontop/alwaysontop.js
```diff
@@ -5,7 +5,8 @@ const {
     onload,
     onbeforeunload,
     shouldImplementDrag,
-    getCurrentSize
+    getCurrentSize,
+    dismiss,
```
hristoterezov: redundant `,`

alwaysontop/alwaysontop.js
```diff
@@ -15,6 +16,9 @@ const {
  */
 let initialSize;
 
+
+document.querySelector('.dismiss').addEventListener('click', dismiss);
```
hristoterezov: please add a check for the element

PR #66, opened 2020-02-14
Add screen sharing tracker window

andrei-gavrilescu:
Functionality that will create an always on top window in bottom center of the screen whenever the user start a content sharing session.
The purpose is to let the user know he is sharing his screen.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,5 +1,5 @@
 const RemoteControl = require('./remotecontrol');
-const {setupScreenSharingForWindow, setupScreenSharingMain} = require('./screensharing');
+const { setupScreenSharingForWindow, setupScreenSharingMain } = require('./screensharing');
 const {
     setupAlwaysOnTopRender,
     setupAlwaysOnTopMain
--- a/index.js
+++ b/index.js
@@ -1,5 +1,5 @@
 const RemoteControl = require('./remotecontrol');
-const setupScreenSharingForWindow = require('./screensharing');
+const {setupScreenSharingForWindow, setupScreenSharingMain} = require('./screensharing');
 const {
     setupAlwaysOnTopRender,
     setupAlwaysOnTopMain
@@ -17,6 +17,7 @@ module.exports = {
     getWiFiStats,
     RemoteControl,
     setupScreenSharingForWindow,
+    setupScreenSharingMain,
     setupAlwaysOnTopRender,
     setupAlwaysOnTopMain,
     setupPowerMonitorRender,
--- a/index.js
+++ b/index.js
@@ -1,5 +1,5 @@
 const RemoteControl = require('./remotecontrol');
-const { setupScreenSharingForWindow, setupScreenSharingMain } = require('./screensharing');
+const { setupScreenSharingRender, setupScreenSharingMain } = require('./screensharing');
 const {
     setupAlwaysOnTopRender,
     setupAlwaysOnTopMain
@@ -16,7 +16,7 @@ const {
 module.exports = {
     getWiFiStats,
     RemoteControl,
-    setupScreenSharingForWindow,
+    setupScreenSharingRender,
     setupScreenSharingMain,
     setupAlwaysOnTopRender,
     setupAlwaysOnTopMain,
--- a/screensharing/main.js
+++ b/screensharing/main.js
@@ -1,0 +1,112 @@
+const { app, BrowserWindow, ipcMain, screen } = require('electron');
+
+const { SCREEN_SHARE_EVENTS_CHANNEL, SCREEN_SHARE_EVENTS, TRACKER_SIZE } = require('./constants');
+
+/**
+ * Main process component that sets up electron specific screen sharing functionality, like screen sharing
+ * tracker and window selection.
+ * The class will process events from {@link ScreenShareRenderHook} initialized in the renderer, and the
+ * always on top screen sharing tracker window.
+ */
+class ScreenShareMainHook {
+    /**
+     * Create ScreenShareMainHook linked to jitsiMeetWindow.
+     *
+     * @param {BrowserWindow} jitsiMeetWindow - BrowserWindow where jitsi-meet api is loaded.
+     * @param {string} identity - Name of the application doing screen sharing, will be displayed in the
+     * screen sharing tracker window text i.e. {identity} is sharing your screen.
+     */
+    constructor(jitsiMeetWindow, identity) {
+        this._jitsiMeetWindow = jitsiMeetWindow;
+        this._identity = identity;
+        this._onScreenSharingEvent = this._onScreenSharingEvent.bind(this);
+
+        // Listen for events coming in from the main render window and the screen share tracker window.
+        ipcMain.on(SCREEN_SHARE_EVENTS_CHANNEL, this._onScreenSharingEvent);
+
+        // Clean up ipcMain handlers to avoid leaks.
+        app.once('window-all-closed', () => {
+            ipcMain.removeListener(SCREEN_SHARE_EVENTS_CHANNEL, this._onScreenSharingEvent);
+        });
+    }
+
+    /**
+     * Listen for events coming on the screen sharing event channel.
+     *
+     * @param {Object} event - Electron event data.
+     * @param {Object} data - Channel specific data.
+     */
+    _onScreenSharingEvent(event, { data }) {
+        switch (data.name) {
+            case SCREEN_SHARE_EVENTS.OPEN_TRACKER:
+                this._createScreenShareTracker();
+                break;
+            case SCREEN_SHARE_EVENTS.CLOSE_TRACKER:
+                if (this._screenShareTracker) {
+                    this._screenShareTracker.close();
+                    this._screenShareTracker = undefined;
+                }
+                break;
+            case SCREEN_SHARE_EVENTS.STOP_SCREEN_SHARE:
+                this._jitsiMeetWindow.webContents.send(SCREEN_SHARE_EVENTS_CHANNEL, { data });
+                break;
+            default:
+                console.warn(`Unhandled ${SCREEN_SHARE_EVENTS_CHANNEL}: ${data}`);
+        }
+    }
+
+    /**
+     * Opens an always on top window, in the bottom center of the screen, that lets a user know
+     * a content sharing session is currently active.
+     *
+     * @return {void}
+     */
+    _createScreenShareTracker() {
+        if (this._screenShareTracker) {
+            return;
+        }
+
+        // Display always on top screen sharing tracker window in the center bottom of the screen.
+        let display = screen.getPrimaryDisplay();
+        this._screenShareTracker = new BrowserWindow({
+            height: TRACKER_SIZE.height,
+            width: TRACKER_SIZE.width,
+            x:(display.bounds.width - TRACKER_SIZE.width) / 2,
+            y:display.bounds.height - TRACKER_SIZE.height - 10,
+            transparent: true,
+            minimizable: true,
+            maximizable: false,
+            resizable: false,
+            alwaysOnTop: true,
+            fullscreen: false,
+            fullscreenable: false,
+            skipTaskbar: true,
+            frame: false,
+            show: false,
+            webPreferences: {
+                nodeIntegration: true
+            }
+        });
+
+        // Prevent newly created window to take focus from main application.
+        this._screenShareTracker.once('ready-to-show', () => {
+            if (this._screenShareTracker && !this._screenShareTracker.isDestroyed()) {
+                this._screenShareTracker.showInactive();
+            }
+        });
+
+        this._screenShareTracker.sharingIdentity = this._identity;
+        this._screenShareTracker.loadURL(`file://${__dirname}/screenSharingTracker.html?`);
+    }
+}
+
+/**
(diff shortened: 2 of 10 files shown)


damencho:
Can you add a screenshot, please :)?

andrei-gavrilescu:
Sure thing.
image

damencho:
Nice, thank you :)

screensharing/main.js
```diff
@@ -0,0 +1,113 @@
...
+ */
+class ScreenShareMainHook {
+    /**
+     * Create ScreenShareMainHook linked to jitsiMeetWindow.
+     *
+     * @param {BrowserWindow} jitsiMeetWindow - BrowserWindow where jitsi-meet api is loaded.
+     * @param {string} identity - Name of the application doing screen sharing, will be displayed in the
+     * screen sharing tracker window text i.e. {identity} is sharing your screen.
+     */
+    constructor(jitsiMeetWindow, identity) {
```
hristoterezov: I'm wondering if can get the `identity` value from electron directly. 

Is `app.name` https://www.electronjs.org/docs/api/app#appname going to do the job?

screensharing/main.js
```diff
@@ -0,0 +1,113 @@
...
+        this._jitsiMeetWindow = jitsiMeetWindow;
+        this._identity = identity;
+        this._onScreenSharingEvent = this._onScreenSharingEvent.bind(this);
+
+        // Listen for events coming in from the main render window and the screen share tracker window.
+        ipcMain.on(SCREEN_SHARE_EVENTS_CHANNEL, this._onScreenSharingEvent);
+
+        // Clean up ipcMain handlers to avoid leaks.
+        app.once('window-all-closed', () => {
```
hristoterezov: I guess even if only `jitsiMeetWindow` is closed we can remove the listener, isn't it? If another one is opened the consumer of the API will anyway need to call all setup methods in order for the screen sharing to work.

Also do we close the screen share tracker if `jitsiMeetWindow` is closed. I think we do because we will receive `_willDispose` event from the iframe api, but can you confirm please?

hristoterezov:
@andrei-gavrilescu A few more things:

- Could you please update the Readme.md?
- Could you please bump the version in package.json? I guess this will be a major bump because this is a braking change for the screensharing APIs.
- Since this is a braking change, could you please fix the screensharing for jitsi-meet-electron?

screensharing/render.js
```diff
@@ -0,0 +1,164 @@
...
+class ScreenShareRenderHook {
+    /**
+     * Creates a ScreenShareRenderHook hooked to jitsi meet iframe events.
+     *
+     * @param {iframe} iframe - The jitsi Meet iframe object. This may seem redundant as we can access it from
+     * the api object, however we need it here for backwards compatibility as some clients are setup is such a
+     * way that they may end up calling the old jitsi-meet-electron-utils with the updated api, causing crashes.
+     * @param {JitsiIFrameApi} api - The Jitsi Meet iframe api object.
+     */
+    constructor(iframe, api) {
```
hristoterezov: `iframe` is kind of redundant since you can get it from `api`

index.js
```diff
@@ -1,5 +1,5 @@
 const RemoteControl = require('./remotecontrol');
-const setupScreenSharingForWindow = require('./screensharing');
+const {setupScreenSharingForWindow, setupScreenSharingMain} = require('./screensharing');
```
hristoterezov: style: space after `{` and before `}`

screensharing/screenSharingTracker.html
```diff
@@ -0,0 +1,71 @@
...
+</head>
+
+<body>
+  <span id="double-line">‖</span>
+  <div id="text-container">
+    <span id="sharing-identity"></span><span id="static-text">&nbsp;is sharing your screen.</span>
+  </div>
+  <div>
+    <button id="screen-share-marker-stop">Stop sharing</button>
+    <a id="screen-share-marker-minimize">Hide</a>
```
hristoterezov: Could you please confirm that it is fine to have not translatable texts here?

screensharing/main.js
andrei-gavrilescu: You could but I prefer for it to be passed down as a parameter, whoever uses this can pass app.name if that's what they need. You would avoid cases where the app.name is very long and maybe not appropriate for the tracker window i.e. (8x8 Free Video Meetings or some other long name).

index.js
andrei-gavrilescu: fixed

screensharing/main.js
andrei-gavrilescu: Indeed it does make more sense to remove the listener when jitsiMeetWindow is closed, will do that.
Yes the screen share tracker is removed when jitsiMeetWindow is closed due to the conference ending event or the _willDispose event from the iframe.

screensharing/main.js
```diff
@@ -0,0 +1,113 @@
...
+        });
+
+        // Prevent newly created window to take focus from main application.
+        this._screenShareTracker.once('ready-to-show', () => {
+            if (this._screenShareTracker && !this._screenShareTracker.isDestroyed()) {
+                this._screenShareTracker.showInactive();
+            }
+        });
+
+        this._screenShareTracker.sharingIdentity = this._identity;
```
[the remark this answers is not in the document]
andrei-gavrilescu: Well, looked around and it seems to be an ok way to pass parameters.

screensharing/main.js
```diff
@@ -0,0 +1,113 @@
...
+        // eslint-disable-next-line no-undef
+        this._screenShareTracker.loadURL(`file://${__dirname}/screenSharingTracker.html?`);
```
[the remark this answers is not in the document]
andrei-gavrilescu: aa neat, added.

screensharing/render.js
andrei-gavrilescu: It is, i left a comment in the parameter section explaining why it's necessary. Basically jitsi-meet-electron-utils is linked to 8x8-meet-electron (standalone) and VOD electron apps, however the code calling it is deployed separately (8x8.vc) and loaded in electron via (iframe), which means that once we make the changes to 8x8.vc to call `new ScreenShareMainHook(api)` instead of new `ScreenShareMainHook(iframe)`  all the old apps (standalone and vod) will load these new changes in the frame which will result in the iframe calling older versions of jtisi-meet-electron-utils inadvertently breaking it.
Long story short we can't change the api.

screensharing/screenSharingTracker.html
andrei-gavrilescu: Honestly I haven't considered that, as it stands I didn't find any support for translating in the project i assumed it was just accepted as is.
I'm guessing changing these strings to parameters would do the trick. So that each application can just pass whatever they want, what do you think?

screensharing/screenSharingTracker.js
```diff
@@ -0,0 +1,30 @@
+const {ipcRenderer, remote} = require('electron');
```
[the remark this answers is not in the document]
andrei-gavrilescu: adde

screensharing/screenSharingTracker.js
```diff
@@ -0,0 +1,30 @@
...
+
```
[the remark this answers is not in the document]
andrei-gavrilescu: don't we delimit library imports from local imports?

screensharing/screenSharingTracker.js
```diff
@@ -0,0 +1,30 @@
...
+const {SCREEN_SHARE_EVENTS_CHANNEL, SCREEN_SHARE_EVENTS} = require('./constants');
```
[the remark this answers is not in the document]
andrei-gavrilescu: added

screensharing/screenSharingTracker.js
```diff
@@ -0,0 +1,30 @@
...
+
+
```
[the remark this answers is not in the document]
andrei-gavrilescu: removed

screensharing/screenSharingTracker.js
```diff
@@ -0,0 +1,30 @@
...
+const screenShareStop = document.getElementById("screen-share-marker-stop");
+const screenShareMinimize = document.getElementById("screen-share-marker-minimize");
+const sharingIdentity = document.getElementById("sharing-identity");
+
+sharingIdentity.innerHTML = remote.getCurrentWindow().sharingIdentity;
+
+/**
+ * Minimize the window.
+ */
+screenShareMinimize.addEventListener("click",function() {
```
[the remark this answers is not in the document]
andrei-gavrilescu: added

screensharing/screenSharingTracker.js
```diff
@@ -0,0 +1,30 @@
...
+    remote.BrowserWindow.getFocusedWindow().minimize();
```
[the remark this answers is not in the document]
andrei-gavrilescu: It's for a button on the right side of the window (the Hide button) that once clicked should minimise the window, as it is on Chrome.

screensharing/screenSharingTracker.js
```diff
@@ -0,0 +1,30 @@
...
+});
+
+/**
+ * When the user clicks the stop button, send a message that will eventually be processed by
+ * {@link ScreenShareRenderHook} which will toggle the screen sharing session using the jitsi-meet api.
+ */
+screenShareStop.addEventListener("click",function() {
```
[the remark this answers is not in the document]
andrei-gavrilescu: added

screensharing/main.js
hristoterezov: OK. That's fine!

screensharing/main.js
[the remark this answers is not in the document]
hristoterezov: But is it documented somewhere? 
Do you know if those custom props are not lost internally at some point?
Does it work on all electron versions?

NOTE: IMHO if it is not documented I'm worried that this may stop working at some point or that it is not always working even now. Also on the top of my head you can very easily pass it trough the IPC objects or define it as a global even. Probably there are even better ways to pass it that match the official electron APIs.

andrei-gavrilescu:
> @andrei-gavrilescu A few more things:
> 
> * Could you please update the Readme.md?
> * Could you please bump the version in package.json? I guess this will be a major bump because this is a braking change for the screensharing APIs.
> * Since this is a braking change, could you please fix the screensharing for jitsi-meet-electron?

I'll update the Readme.md.
You mean bump the version in jitsi-meet-electron and update the code there to use the new stuff?

screensharing/screenSharingTracker.js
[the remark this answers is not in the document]
hristoterezov: Oh ok. Just tested Chrome and it really minimizes it. I was thinking it is actually closing it :)

screensharing/screenSharingTracker.js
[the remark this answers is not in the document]
hristoterezov: Hmm.. We actually don't have strict policy in this project about that. But I like the intent.

hristoterezov:
> > @andrei-gavrilescu A few more things:
> > 
> > * Could you please update the Readme.md?
> > * Could you please bump the version in package.json? I guess this will be a major bump because this is a braking change for the screensharing APIs.
> > * Since this is a braking change, could you please fix the screensharing for jitsi-meet-electron?
> 
> I'll update the Readme.md.
> You mean bump the version in jitsi-meet-electron and update the code there to use the new stuff?

You definitely need to update the code in jitsi-meet-electron to use the new stuff. But also you need to bump the version in jitsi-meet-electron-utils's package.json. Once we merge the PR we will create a new release and you will be able to update jitsi-meet-electron's package.json to use the new jitsi-meet-electron-utils.

screensharing/screenSharingTracker.html
hristoterezov: Yes we don't have translation support here. Passing the strings is too much customisation for me :) Also it won't be so easy because you will have to add support for changing them realtime for the use case when the user changes the language. Let's take this offline and decide what to do!

screensharing/render.js
hristoterezov: OK. I see. I'm sorry I missed the comment. Let me thing about it and discuss offline.

hristoterezov: OK. I see. I'm sorry I missed the comment. Let me thing about it and discuss offline.

screensharing/main.js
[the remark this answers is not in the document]
andrei-gavrilescu: changed to o more accepted approach.

andrei-gavrilescu:
> > > @andrei-gavrilescu A few more things:
> > > 
> > > * Could you please update the Readme.md?
> > > * Could you please bump the version in package.json? I guess this will be a major bump because this is a braking change for the screensharing APIs.
> > > * Since this is a braking change, could you please fix the screensharing for jitsi-meet-electron?
> > 
> > 
> > I'll update the Readme.md.
> > You mean bump the version in jitsi-meet-electron and update the code there to use the new stuff?
> 
> You definitely need to update the code in jitsi-meet-electron to use the new stuff. But also you need to bump the version in jitsi-meet-electron-utils's package.json. Once we merge the PR we will create a new release and you will be able to update jitsi-meet-electron's package.json to use the new jitsi-meet-electron-utils.

Got it, bumped the version.

saghul:
> Since this is not backward-compatible change the version needs to be 2.0.0, isn't it?

Yep.

andrei-gavrilescu:
woops, changing.

PR #67, opened 2020-03-05
Fix screen-share tracker on Windows 10 OS

andrei-gavrilescu:
On Windows we need to specify `-webkit-app-region: no-drag;` for elements of a draggable window in order to interact with them, i.e. the Stop sharing button in our case.
Also from time to time some sort of provisory font is loaded before the css specified one, which caused text shift one line, giving the window more width seems to fix this problem.

### DIFF
--- a/screensharing/main.js
+++ b/screensharing/main.js
@@ -89,8 +89,6 @@ class ScreenShareMainHook {
             }
         });
 
-        this._screenShareTracker.openDevTools();
-        
         // Prevent newly created window to take focus from main application.
         this._screenShareTracker.once('ready-to-show', () => {
             if (this._screenShareTracker && !this._screenShareTracker.isDestroyed()) {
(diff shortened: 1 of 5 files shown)


screensharing/main.js
```diff
@@ -72,16 +72,16 @@ class ScreenShareMainHook {
...
+            y:display.workArea.height - TRACKER_SIZE.height - 10,
             transparent: true,
             minimizable: true,
             maximizable: false,
             resizable: false,
             alwaysOnTop: true,
             fullscreen: false,
             fullscreenable: false,
-            skipTaskbar: true,
+            skipTaskbar: false,
```
saghul: why is this change needed?

andrei-gavrilescu: On windows if you click hide on screen share tracker, it won't appear as a minimised window in the task bar, so basically you won't be able to bring it up.
Chrome does the same thing with their screen share track window.

andrei-gavrilescu:
> LGTM with a question. Please squash when merging.

Roger.

screensharing/main.js
saghul: Makes sense, thanks for explaining!

andrei-gavrilescu: On mac this doesn't seem to have any effect.

PR #72, opened 2020-04-06
Keep AoT window open during meeting: hide & show insetad of close & open

andreieftimie:
Workaround for electron crashing (strictly a crash, but a failure that ends in the app being forcefully closed) when AoT window is quickly closed and opened during both a focus (main window) and navigate away from the meeting page on Windows.

I would also only hide on dbclick but this breaks current integration as navigating is only one on close.

I also have some performance concerns as the AoT window will still be open and video is piped to it even when it's not visible, I haven't deeply looked into also tackling this potential perf issue.

### DIFF
--- a/alwaysontop/render.js
+++ b/alwaysontop/render.js
@@ -8,6 +8,16 @@ const path = require('path');
 const { ALWAYSONTOP_DISMISSED, ALWAYSONTOP_WILL_CLOSE, SIZE } = require('./constants');
 
 /**
+ * Checks if a BrowserWindow object is defined and that it's not destroyed
+ *
+ * @param  {BrowserWindow} win - electron BrowserWindow instance to be checked
+ * @return {boolean} - whether the window object exists and can be worked with
+ */
+function exists(win) {
+  return win && !win.isDestroyed();
+}
+
+/**
  * Returieves and trying to parse a numeric value from the local storage.
  *
  * @param {string} localStorageKey - The key of the value that has to be
@@ -47,6 +57,8 @@ class AlwaysOnTop extends EventEmitter {
         this._updateLargeVideoSrc = this._updateLargeVideoSrc.bind(this);
         this._openAlwaysOnTopWindow = this._openAlwaysOnTopWindow.bind(this);
         this._closeAlwaysOnTopWindow = this._closeAlwaysOnTopWindow.bind(this);
+        this._showAlwaysOnTopWindow = this._showAlwaysOnTopWindow.bind(this);
+        this._hideAlwaysOnTopWindow = this._hideAlwaysOnTopWindow.bind(this);
         this._onMessageReceived = this._onMessageReceived.bind(this);
         this._onConferenceJoined = this._onConferenceJoined.bind(this);
         this._onConferenceLeft = this._onConferenceLeft.bind(this);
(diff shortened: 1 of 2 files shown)


andreieftimie:
@hristoterezov please review to see if this strategy is fine, or propose alternatives. We used a patched version of `jitsi-meet-electron-utils` (with these changes on top of `1.1.3`) to enable a current release, but going forward I'd like to use an official release

alwaysontop/render.js
```diff
@@ -7,6 +7,10 @@ const path = require('path');
 
 const { ALWAYSONTOP_DISMISSED, ALWAYSONTOP_WILL_CLOSE, SIZE } = require('./constants');
 
+function exists(win) {
```
hristoterezov, line 10: Please add JSDoc.

alwaysontop/render.js
```diff
@@ -367,7 +374,8 @@ class AlwaysOnTop extends EventEmitter {
      * @returns {void}
      */
     _closeAlwaysOnTopWindow() {
-        if (this._alwaysOnTopBrowserWindow && !this._alwaysOnTopBrowserWindow.isDestroyed()) {
+
```
hristoterezov, line 377: style: redundant empty line

alwaysontop/render.js
andreieftimie: Done

andreieftimie: Done

alwaysontop/render.js
andreieftimie: Removed.

andreieftimie:
> Could you please provide more details about the issue that you are fixing?

As noted in the description, there's an electron "crash" (doesn't actually register as a crash, but behaves like one). 

This is the error thrown:

```
Error Object has been destroyed 
    electron/js2c/browser_init.js:535:47 WebContents.
    events.js:203:13 WebContents.emit
    domain.js:476:20 WebContents.EventEmitter.emit
    domain.js:131:23 WebContents.topLevelDomainCallback
```

This code is only executed on Windows (or at least not executed on macOS, have not tested linux).

I can reproduce with the following steps:

1. enter a meeting
2. click on menu Meeting icon (show the meeting iframe)
    => AoT window is closed
3. click outside of the app window
    => AoT window is opened due to blur
4. click on menu Message icon
    => this triggers 2 separate flows which is the heart of the problem: 
    a. focus on the app window wants to hide the AoT window
    b. in-app navigation away from the meeting view (iframe with the actual meeting) wants to show the AoT window

This is highly intermittent, run steps 2-4 in a loop (I need 5-50 tests to reproduce it). It always fails on step 4

I was not able to precisely pinpoint where things go wrong. It might also be an internal electron issue. 

The internal electron code is the following:

```
  // Hide the auto-hide menu when webContents is focused.
  this.webContents.on('activate', () => {
    if (process.platform !== 'darwin' && this.autoHideMenuBar && this.isMenuBarVisible()) {
      this.setMenuBarVisibility(false)
    }
  })
```
https://github.com/electron/electron/blob/7-1-x/lib/browser/api/browser-window.js#L36

Basically an `activate` event is triggered from a `webContents` instance, but when this code attempt to run the `BrowserWindow` object no longer exists — when this failure happens `this` in the above code points to the destroyed object and fails.

I'll PM you the internal ticket for more details.

-----

> Why we can't use this approach for double click?

I couldn't use this as a quick fix due to the nature of the jitsi integration. It can made to work, but requires additional changed (in our internal implementation) that requires precise coordination.

1. double click on the AoT window does not bring the meeting page into view
2. this is because the navigation to the meeting page is done only when the AoT window closes in the deployed meeting code
3. this means I have to still close the window on double click, which does bring the meeting page into view, but adds noticeable delay when the AoT window is opened again
4. my concern is that this might bring the crash back, but so far I have not been able to reproduce but I still have my concerns

I'll include internal details in the PM.

-----

> What is the performance impact? Have you tested?

I have not had the chance to verify the performance hit. I would think there is one as we're still piping the video to a hidden window.

> Maybe we can improve it by removing the video while the AOT is hidden and stop updating it. WDYT?

This was my though as well, though I haven't looked into this at the moment — this was my first foray into this codebase and had to address the "crash" in a very tight schedule. Not sure if / when I'll have time to address vide piping.

PR #73, opened 2020-04-07
alwaysontop/renderer(fix): Prevent crash on new browser window

horymury:
@saghul @hristoterezov 

Encountered this bug where the browserwindow was not found based on the id. Added a failsafe mechanism for this case.


hristoterezov:
@horymury I'm wondering if we need to clear some more things there in order to not be in some weird state. When is this happening? Have we called window.open? Do we have value for `this._alwaysOnTopBrowserWindow` when this happens?

horymury:
@hristoterezov I could not figure out exactly when this happens, the error was reported by other people and I was not able to repro. I think once we'll have some logging to AOT we could tell what went wrong. I could deal with adding logging.

PR #81, opened 2020-04-28
Fix for shiboleth authentication with CSP

immerda:
This commit removes CSP frame-anchestor and x-frame-option from
response headers. This prevents electron from blocking server side
sign-in (such as Shiboleth/SAML authentication) with those headers.

### DIFF
--- a/popupsconfig/main.js
+++ b/popupsconfig/main.js
@@ -14,6 +14,26 @@ function initPopupsConfiguration(jitsiMeetWindow) {
         popupsConfigRegistry.registerPopupConfigs(configs);
     });
 
+    // Remove X-Frame-Option and CSP frame-anchestor headers from responses, to
+    // allow us to show login pages from the server, even if it sets those
+    // headers. This is for example often needed for Shiboleth authentication.
+    var onHeadersReceived=(detail, callback) => {
+        const xFrameOriginKey = Object.keys(detail.responseHeaders).find(
+          header => String(header).match(/^x-frame-options$/i));
+        const cspKey = Object.keys(detail.responseHeaders).find(
+          header => String(header).match(/^content-security-policy$/i));
+        if (xFrameOriginKey) {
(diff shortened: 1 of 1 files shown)


saghul:
Very clever solution! Alas I don't think this is a good idea without an allowlist.

immerda:
@saghul good point to be vigilant here. but what are the concerns here? There are two issues:

* danger to jitsi-meet: that is not an issue. if a malicious site wants to display something inside jitsi-meet it will just not set the no-frame header
* danger to the site: indeed jitsi-meet would violate the choice of a website to have it not framed. however i would argue that what jitsi-meet does is not framing, since effectively it displays the login form full-screen. also currently the inverse danger is even bigger. we had to remove the "no-frame" settings from our login page to make it compatible with jitsi-meet.

Overall personally I find ignoring the header is the lesser of the two evils.

However if you find that a hard-coded whitelist, where service providers can just create a PR here to get included is appropriate, then I can certainly change the PR to do that.

immerda:
@saghul any news here? would you want us to implement an allowlist?

EnercoopJNE:
Hi ! Is there something preventing merge ? The issue is still there : can't use SAML with the client...

saghul:
Looks like we dropped the ball here :-/ I'll se another look.

EnercoopJNE:
Merging is still needing a review apparently...

immerda:
@saghul Sorry, but doesn't this issue also apply to non-shibboleth problems, like the JWT one.

If I look at the flow in the linked saml2jwt repo from the linked thread: https://github.com/Renater/Jitsi-SAML2JWT You will see that there will still be some dance the jitsi client will have to do and thus the restriction still applies.

Or how do you think the jtisi client will not have to embed any login websites anymore when going with the jwt solution? The only thing you are changing is the kind of assertion you need to support within jitso. But the client will still need to display a login page to obtain the jwt.

PR #82, opened 2020-05-04
fix(AOT): will close event.

hristoterezov:
@andreieftimie After merging https://github.com/jitsi/jitsi-meet-electron-utils/pull/72 I realised that actually we have broken the will-close event. This PR brings back the original behaviour of the event. Since it is abstract event I'm triggering it for the use cases when we hide the AOT. In addition I realised that this would probably solve the issue you described for the double click. That's why I also changed the double click to hide AOT instead of close it.

Could you please verify that this will work for your use case?

### DIFF
--- a/alwaysontop/main.js
+++ b/alwaysontop/main.js
@@ -86,7 +86,7 @@ function onAlwaysOnTopWindow(
 
         win.webContents.on('error', error => {
             console.warn(error, 'Unhandled AOT webContents error');
-        })
+        });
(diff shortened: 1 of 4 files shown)


saghul:
IS this still alive?

andreieftimie:
We're going to try to test this today.

andreieftimie:
@hristoterezov looks good to me. Tested the happy flows.

PR #83, opened 2020-05-19
Fix AoT close race condition

ilazar:
After AoT is closed, calling win methods (e.g. hide) throw exceptions even the isDestroyed method returns false.


saghul:
When are you seeing exceptions?

ilazar:
This module is used by 8x8 Virtual Office, and even with the fix which checks whether win isDestroyed we saw crashes with errors like

"Could not call remote method 'hide'. Check that the method signature is correct. Underlying error: Object has been destroyed"

for all methods getPosition, etc.

PR #92, opened 2020-09-07
Use AOT window layering to allow it to render above full screen apps

andreieftimie:
This solves AOT rendering above full-screen applications (on windows Remote Desktop Connection for example) which allow users in a meeting to open a RDP connection and still see the meeting AOT window above it (especially important if the user has a single screen).

See a lengthy in-depth discussion here https://github.com/electron/electron/issues/10078

### DIFF
--- a/alwaysontop/main.js
+++ b/alwaysontop/main.js
@@ -70,6 +70,9 @@ function onAlwaysOnTopWindow(
             }, getPosition(), getSize())
         );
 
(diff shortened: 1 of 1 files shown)


andreieftimie:
Ping @hristoterezov

andreieftimie:
Can we get this in?

saghul (approved):

PR #94, opened 2020-10-06
[AOT] Fix `focus` listener being removed permanently in a certain scenario

horymury:
@saghul @hristoterezov

### DIFF
--- a/alwaysontop/render.js
+++ b/alwaysontop/render.js
@@ -69,12 +69,15 @@ class AlwaysOnTop extends EventEmitter {
         this._onConferenceJoined = this._onConferenceJoined.bind(this);
         this._onConferenceLeft = this._onConferenceLeft.bind(this);
         this._onIntersection = this._onIntersection.bind(this);
+        this._onFocus = this._onFocus.bind(this);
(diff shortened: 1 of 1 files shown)


saghul:
I'm not very familiar with that part of the code, sorry! I leave it up to Hristo here.

hristoterezov (commented):
@horymury Could you please describe the scenario?

This looks a little strange to me. Consider following scenario:
1) The user executes an action in the app that uses the AOT util and this results in hiding the iframe in order to show other functionality.
2) The intersect listener is called and we show the AOT window.
3) The user changes the focus of the app by clicking on another window.
4) The user returns back to the app and the iframe is  still hidden

Before the PR it seems to me that the AOT window will be displayed after 3) and 4) which seems correct to me. But after the PR,  we add the focus listener in the case when the iframe is hidden too, the user won't see the AOT window after 4) because the  focus listener will be executed. This seems incorrect to me.

Am I missing something?

horymury:
@hristoterezov I changed the approach for this fix a bit, I think it's cleaner now.

hristoterezov (approved):

PR #100, opened 2020-12-11
fix(remote-control): some multi monitor use cases

hristoterezov:


remotecontrol/index.js
```diff
@@ -113,7 +113,19 @@ class RemoteControl {
                 } else {
                     // On Mac OS the sourceId = 'screen' + displayId.
                     // Try to match displayId with sourceId.
-                    const displayId = Number(parsedSourceId);
+                    let displayId = Number(parsedSourceId);
+
+                    if (isNaN(displayId)) {
+                        // The source id may have the following format "desktop_id:0".
```
saghul, line 119: Is this the literal string `desktop_id` followed by a colon? If so, why not use replace like above?

saghul (approved):
LGTM with a question / suggestion.

remotecontrol/index.js
hristoterezov: No no, this is the actual id we need to extract in order to match the display against the array returned from `remote.screen.getAllDisplays();`!

PR #101, opened 2021-01-04
Remove mac-screen-capture-permissions

csett86:
mac-screen-capture-permissions prevents successful notarization for Big Sur:
https://github.com/karaggeorge/mac-screen-capture-permissions/issues/5

Screensharing continues to work with Electron 11, the question to enable screen
recording permission is just moved to the first instance of screensharing
instead of first launch.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -759,20 +759,6 @@
         "safer-buffer": "^2.1.0"
       }
     },
-    "electron-is-dev": {
-      "version": "1.2.0",
-      "resolved": "https://registry.npmjs.org/electron-is-dev/-/electron-is-dev-1.2.0.tgz",
-      "integrity": "sha512-R1oD5gMBPS7PVU8gJwH6CtT0e6VSoD0+SzSnYpNm+dBkcijgA+K7VAMHDfnRq/lkKPZArpzplTW6jfiMYosdzw=="
-    },
-    "electron-util": {
-      "version": "0.13.1",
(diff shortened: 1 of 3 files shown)


csett86:
Even with perfectly signed and notarized .app the following error will otherwise appear on Big Sur:

Bildschirmfoto 2021-01-04 um 16 20 55

As the libary is not necessary for successful screenshare, remove it.

csett86:
@saghul any chance to get this reviewed? mac-screen-capture-permissions is blocking both an M1 version of the Jitsi electron app (as the unnecessary dependency contains an x86 binary and no rebuild instructions) and prevents successful notarization check on Big Sur. 

I would prefer to drop my own fork (https://github.com/csett86/jitsi-meet-electron-utils) rather sooner than later :)

saghul:
Hum, the problem I see here is that we are no longer able to reset the permissions so the system prompt won't show again, leaving the users with no clue about what to do. IIRC that was the main reason why we introduced the module.

The proper fix would probably be to build the mac-screen-capture-permissions library with an arm64 slice too, as a "fat binary".

Or drop this library and use some other way to accomplish the same, but as it stands I cannot accept this PR, sorry.

csett86:
I understand and I gave it some more thought. I have replaced the bundled executable in mac-screen-capture-permissions with a n-api native module that does the same thing: https://github.com/karaggeorge/mac-screen-capture-permissions/pull/6

Hopefully this can get merged, then a new release cut for mac-screen-capture-permissions, and then updated here in jitsi-meet-electron-utils.

saghul:
Excellent work!

PR #102, opened 2021-01-09
Prebuild linux, mac and windows binaries

csett86:
update prebuild and add node-abi to be able to build with current
electron and node versions.


csett86:
Just noticed that there are already prebuilds available, closing this....

PR #117, opened 2021-05-13
refactor(aot): replace robotjs with electron screen api

csett86:
robotjs has no support on wayland, replace with electron native api

### DIFF
--- a/alwaysontop/main.js
+++ b/alwaysontop/main.js
@@ -1,6 +1,5 @@
 const os = require('os');
 const electron = require('electron');
-const robot = require('robotjs');
 const { BrowserWindow, ipcMain } = electron;
 const { SIZE } = require('./constants');
 const log = require('jitsi-meet-logger');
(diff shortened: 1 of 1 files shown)


csett86:
@saghul Please review. I have tested this successfully and as robotjs is not context-aware, it is not compatible with Electron 14+, thus replace it as best as possible.

saghul:
Nice! We have 8x8 apps using this package, so we can't just require Electron 14.

I think we could make this the 3.0 release.

@hristoterezov thoughts?

@csett86 I'm quite busy these days, I'll try to get to this PR next week or the one after. I think we also have some remote module usages here that prevent us from using 14, don't we?

csett86:
Yes, you are totally right, the contect-aware topic is not the only (and probably the smaller) issue.

I was planning to tackle the other obstacles for electron 14 in separate PRs as well.

saghul:
> I was planning to tackle the other obstacles for electron 14 in separate PRs as well.

Sounds like a plan!

csett86:
No longer necessary after #129, closing

PR #119, opened 2021-05-23
refactor: replace mac-screen-capture-permissions with electron API

csett86:
The behaviour is exactly the same as before, as the old
hasScreenCapturePermission() would set the "prompt" marker file
that hasPromptedForPermission() would then detect, thus promptedAlready
was always true.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -756,20 +756,6 @@
         "safer-buffer": "^2.1.0"
       }
     },
-    "electron-is-dev": {
-      "version": "1.2.0",
-      "resolved": "https://registry.npmjs.org/electron-is-dev/-/electron-is-dev-1.2.0.tgz",
-      "integrity": "sha512-R1oD5gMBPS7PVU8gJwH6CtT0e6VSoD0+SzSnYpNm+dBkcijgA+K7VAMHDfnRq/lkKPZArpzplTW6jfiMYosdzw=="
-    },
-    "electron-util": {
-      "version": "0.13.1",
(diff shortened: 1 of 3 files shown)


csett86:
@saghul This might be an easier option going forward than getting prebuilt binaries released into mac-screen-capture-permissions (and keeping them up to date with every released electron version)

csett86:
Status update: This is successfully tested on macOS 11.4 and behaves exactly as the the code that it replaces.

@saghul or @hristoterezov, any objections to get this merged?

csett86:
@saghul I successfully tested this on macOS 11.4. Please approve and merge if you agree to the approach.

saghul:
I’ll need some time to dig into this.

csett86:
@saghul Please take a look again at this, this would be much appreciated. Thank you!

saghul:
LGTM, what electron versions are supported?

csett86:
this is supported since electron 8: https://github.com/electron/electron/commit/14cc902ad83a21c1dab7765166e9f3402cb48c3c

saghul:
That's odd, why didn't we use it then?! Maybe it got added later down the line and backported?

csett86:
It got added in 8.0.0 beta3, https://github.com/electron/electron/blob/v8.0.0-beta.3/docs/api/system-preferences.md, so not sure why it was not used back then, but I still suggest to use it now.

saghul (approved):

PR #120, opened 2021-05-26
fix(remotecontrol): make native module context-aware

csett86:
Fixes deprecation warning as all modules have to be context-aware for
electron 14+

See https://github.com/electron/electron/issues/18397


csett86:
@saghul please review, this is required for Electron 14. I have tested this successfully already. Thank you.

saghul (approved):

PR #121, opened 2021-05-27
fix(SS): when the iframe reloads.

hristoterezov:


damencho:
What is the difference with the previous logic? 
Can you add some description what is the problem and how it is fixed ...

andrei-gavrilescu:
Aside from @damencho comment LGTM.

damencho (approved):

PR #129, opened 2021-11-01
refactor(aot): replace remote module for electron 14+

quitrk:
@csett86 rebased with master, might have missed some things

Also, I'll have to undo the remote control removal. Any particular reason you removed it?


csett86:
@quitrk I guess you only want to cherry-pick the your single aot rewrite commit instead of taking all of the other commits here as well.

I have already upstreamed everything that is fine for upstream, the other parts like remote module removal I just have on my own fork to have all native dependencies removed. This is not meant for upstream.

csett86:
I see that the electron-store usage is on main thread, not as the localStorage before on the render thread. With that, electron-store should be fine, but then it needs to be added as a runtime dependency as part of this PR, I guess.

csett86:
But even with that I still get an error when trying to load the aot due to not found electron-store:

electron-store

quitrk:
Added electron-store as peer dependency. It was working for me cause I was testing with jitsi-meet-electron which already has it

csett86 (commented):
Can you quickly explain why you added it as a peer dependency? I would have expected it as a normal dependency, as it does not fit to the description given here: https://nodejs.org/es/blog/npm/peer-dependencies/

quitrk:
I figured I would let the host app decide on the version, since we only do get / set operations which are (most probably) supported on most versions. Was thinking that we could get rid of some bundle size this way, though it might make more sense if we go this route to lower the version requirement

quitrk:
Moved to a normal dependency. lmk if it works for you

alwaysontop/main/utils.js
```diff
@@ -1,129 +1,72 @@
-const os = require('os');
+const store = require('electron-store');
```
csett86, line 1: This does not work for me, I always (on Linux and Mac) get the following error:

aot-electron-store

According to the docs of electron-store, it should be something like:

```
const Store = require('electron-store');

const store = new Store();
```

I'll push this as a fixup commit and test.

csett86 (approved):
Looking good now, I also tested successfully on Linux, Win and Mac. I would squash-merge it, so that the initial commit message and your authorship is clear.

csett86 (approved):

quitrk:
awesome, thank you

PR #130, opened 2021-11-01
ci: move to macOS 11 and node 16

csett86:


csett86:
@saghul, can you please review this as well, to get jitsi-meet-electron and jitsi-meet-electron-utils back on the same node / npm version for CI?

saghul (approved):

Issue #126, opened 2021-09-17
Replace electron remote API to be ready for Electron 14+


csett86:
Electron 14 removes the deprecated remote API: https://www.electronjs.org/docs/breaking-changes#removed-remote-module

At the same time Microsoft requires Chrome-based apps (including Electron) to be not older than 2 major versions for the Microsoft Store: https://www.electronjs.org/blog/8-week-cadence/ (in effect since July 28th)

Thus the goal to remove or replace all remote API calls from jitsi-meet-electron-utils.

@electron/remote is not a drop-in replacement, as it cannot be "required" on the main thread, thus it needs to be carefully "required" only inside functions that are on the render thread (via @electron/remote) and on the main thread (via @electron/remote/main).

csett86:
I will work on this topic as I maintain https://github.com/HoppFoundation/jitsi-meet-electron which is published via the Microsoft Store.

csett86:
WIP is tracked here: https://github.com/csett86/jitsi-meet-electron-utils/tree/remote

PR #134, opened 2021-11-03
BREAKING CHANGE: replace remote module in remotecontrol for electron 14+

csett86:
Now requires are component running on main process, thus breaking existing API, thus major version bump included as well.

Closes: #126


csett86:
I already tested it successfully on Linux and Mac, Windows testing in progress

csett86:
@saghul, @quitrk: Successfully tested on Windows, Linux and Mac, please review together with https://github.com/jitsi/jitsi-meet-electron/pull/655

saghul (approved):
LGTM. I see we still have robotjs in there, wasn't that going away too?

csett86:
aot had a single use of robotjs that could be easily replaced, but remotecontrol unfortunately uses a lot more apis of robotjs. And its not strictly necessary to replace, as it works with electron 14+ as long as it is made context-aware.

PR #135, opened 2021-11-04
remotecontrol: Move to context-aware robotjs fork for electron 14+

csett86:
Will be reverted as soon as https://github.com/octalmage/robotjs/pull/696
is merged upstream.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6,7 +6,7 @@
   "packages": {
     "": {
       "name": "jitsi-meet-electron-utils",
-      "version": "2.0.22",
+      "version": "3.0.0",
       "hasInstallScript": true,
       "license": "Apache-2.0",
       "dependencies": {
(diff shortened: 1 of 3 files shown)


package-lock.json
```diff
@@ -1,4884 +1,8 @@
 {
   "name": "jitsi-meet-electron-utils",
   "version": "3.0.0",
-  "lockfileVersion": 2,
```
saghul, line 4: Please undo this change, aka, run npm install with npm 8

saghul:
LGTM, minue the downgrade in package-lock

csett86:
Agree, the downgrade was due to the #130 not being merged… will redo package-lock update later.

package-lock.json
csett86: Done

csett86:
Now with npm8, please review

saghul (approved):
LGTM, let's hope the failure was a fluke,  I retried the job.

csett86:
I am also locally checking if I can reproduce it in a win10 VM... To me the build failure looks reproducible, and the strange thing is that with npm6 and node12 it was working, and on all other platforms its also working...

csett86:
I finally found and fixed the error: prebuild has a dependency on node-gyp, which was then too old and failing with the non-descript erros. Updating prebuild (in the robotjs fork) helped.

saghul:
Wonderful!

saghul (approved):

PR #139, opened 2021-11-09
Prepare for publishing prebuilt artifacts for Electron 14

saghul:


csett86:
Before this can go in, https://github.com/prebuild/prebuild-install/pull/161 has to be merged, released and we need to use here. Otherwise prebuild-install will download the artefacts with the wrong ABI version.

This is also the reason why I removed v89 artefacts from the 3.0.0 prerelease for the time being.

csett86:
In addition, please use electron 14.1.1 as 14.0.0 has wrong ABI version 89 in the metadata: https://github.com/electron/releases

saghul:
Good to know. This will only kick in when we actually release 3.0.0, and I don't think we are quite there yet, are we?

csett86:
3.0.0 works fine with electron 14 in my local tests, I dont see any blockers left. Thats why I have at least published it as prerelease and jitsi-meet-electron is using it (but not yet in a released version).

The only issue I know of are the ABI versions of the prebuilds during install, do you have other open topics in mind?

saghul:
> do you have other open topics in mind?

I opened a couple of issues here to track them. The robotjs one is a blocker because it would affect other teams at 8x8 who use this package, their build pipeline would be affected, but providing prebuilts would solve that easily. The module thing is more of a nice to have.

csett86:
Instead of having to deal with the node abi versions, I have refactored the native code to use the stable ABI Node-API: https://github.com/jitsi/jitsi-meet-electron-utils/pull/141

Please review there, this obsoletes the continuous prebuilds for new electron versions (espectially as electron has shortened the release schedule from 12 to 8 weeks per major version).

PR #142, opened 2021-11-10
BREAKING CHANGE: rename package to @jitsi/electron-sdk

saghul:


csett86 (approved):
Looks good to me, I assume npm publish will follow under this name then as well?

saghul:
> Looks good to me, I assume npm publish will follow under this name then as well?

Correct.

PR #146, opened 2021-11-11
ci: publish release to npm

### LINKED ISSUE #144: Automate publishing to NPM with GH actions
adding the npm registry to the `setup-node` step and a 

    npm publish

with npm token via secret

    NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}

and guarded with the release published conditions should do it according to: https://docs.github.com/en/actions/publishing-packages/publishing-nodejs-packages#publishing-packages-to-the-npm-registry

csett86:
Requires npm access token in secret `NPM_TOKEN`

Closes #144

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,6 +19,7 @@ jobs:
     - uses: actions/setup-node@v2
       with:
         node-version: '16'
+        registry-url: 'https://registry.npmjs.org'
     - name: Install dependencies
       if: startsWith(matrix.os, 'ubuntu')
       run: sudo apt-get install libxtst-dev libpng++-dev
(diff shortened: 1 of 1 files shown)


saghul:
We already have a PUBLISH_TOKEN secret. Let's try that?

csett86:
The PUBLISH_TOKEN is a github API token, I would assume (as its used to publish to GH releases), here we need a npmjs access token. Even though npmjs is nowadays owned by github, I could not find any reference that github tokens works to publish to npm. Did I miss anything?

saghul:
Ah wait, you're right.

saghul:
Done.

saghul (approved):

PR #154, opened 2021-11-16
chore(deps): update robotjs to use prebuilds during npm install

csett86:


saghul (approved):

saghul:
Shall we make a .1 release now?

csett86:
Have you had any further input from 8x8 internal groups or team on 3.0.0 any any issues? If yes, I would fix them and then do the 3.0.1, if there is no feedback (yet), then I would directly do the 3.0.1

saghul:
No feedback yet, I'd so a 3.0.1 so we don't run into prebuilt issues.

csett86:
Agree, then here we go: https://github.com/jitsi/jitsi-meet-electron-sdk/pull/155

PR #162, opened 2021-12-11
cleanup: remove unused node-abi

csett86:
node-abi was introduced to get prebuilds with "prebuild" with
current electron versions. "prebuild" is replaced with "prebuildify",
thus node-abi is no longer used.


csett86:
@saghul can you please review this cleanup? As its just removing an unused dependency that I missed when switching from prebuild to prebuildify.

saghul (approved):

PR #163, opened 2021-12-11
chore(deps): enable dependabot

csett86:


csett86:
@saghul, do you agree to enable dependabot here? If so, please approve :)

saghul (approved):

saghul:
Sure, let's have it!

PR #165, opened 2021-12-14
fix(aot-breakout) Prevent multiple calls to join and leave meet handlers

horymury:


csett86:
@horymury, are you sure you want to merge this into revert-mac-capture? I would assume master should be the target of this PR?

horymury:
> @horymury, are you sure you want to merge this into revert-mac-capture? I would assume master should be the target of this PR?

This is an internal branch, I'll put the fix also on the master branch tomorrow as the issue reproduces there as well.

PR #166, opened 2021-12-15
fix(aot-breakout) Prevent `videoConferenceJoined` handler being calle…

horymury:
…d multiple times

- When joining breakout rooms, a new `videoConferenceJoined` is triggered. Also yet another `videoConferenceJoined`
is triggered when leaving the breakout room. This seem to cause at least the onBlur handler not to be removed,
thus onBlur gets called after the meeting was left, which creates a new AOT window.


quitrk (approved):

quitrk (changes requested):
let's add a discriminator for breakout rooms on leave / join events and just ignore them based on that

damencho:
Hey @horymury the problem with AOT window is that we do not fore videoConferenceLeft event when moving between main room and breakout rooms. We have a fix for that: https://github.com/jitsi/jitsi-meet/pull/10627
Can you test with that PR that everything works as expected?

PR #173, opened 2022-01-10
fix(aot) Fix intersection; fix conference left

horymury:
- Now focus handler also checks whether the meeting is intersecting before hiding AOT
- Now on conference left we do not remove the listeners because it not always means that meeting ended.
This fixes cases when we have a passcode/lobby and also breakout rooms when conferenceLeft and joined are called
during the same meeting session.


quitrk:
did you test this with the various integrations we have for it?

horymury:
> did you test this with the various integrations we have for it?

yep I did.

PR #180, opened 2022-01-19
fix(aot) Fix camera preview becoming unsynced when screen sharing.

horymury:
- fix camera preview sometimes showing as a black screen in the bottom left corner when
screen sharing
- fix camera preview sometimes being out of sync with the in-meeting video off/on state
and instead showing a black video when video is off and no video when video is on


csett86 (approved):
I assume you have tested this

horymury:
> I assume you have tested this

Yep, and I'll test some more before creating a new release. Thanks for the approve.

PR #186, opened 2022-01-29
fix(screenshare): adapt to electron 17+ desktopCapturer.getSources

csett86:
Starting with electron 17, desktopCapturer.getSources(options) is
only available on the main process, not the renderer.

See https://www.electronjs.org/docs/latest/breaking-changes#removed-desktopcapturergetsources-in-the-renderer

### DIFF
--- a/screensharing/constants.js
+++ b/screensharing/constants.js
@@ -5,6 +5,12 @@
 const SCREEN_SHARE_EVENTS_CHANNEL = 'jitsi-screen-sharing-marker';
 
 /**
+ * The name of the channel that returns desktopCapturer.getSources
+ * @type {string}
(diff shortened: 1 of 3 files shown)


csett86:
@saghul please review, I have successfully tested this, the deprecation warning is gone and the list of sources (eg screens and applications) still work.

csett86:
@saghul, does it look fine for you? I have already tested it successfully

saghul (approved):

saghul:
Sorry I missed this. LGTM!

csett86:
All good, thanks for the review. Its not that urgent. I will cut a release afterwards to have this ready for use.

PR #210, opened 2022-03-15
Darwin x64 only

saghul:
Otherwise projects not building for arm64 on macOS (yet) run into a build problem because the prebuild cannot be found.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,6 +20,9 @@ jobs:
           - os: macos-11
             arch: x64
             build-group: darwin-universal
+          - os: macos-11
+            arch: x64
+            build-group: darwin-x64
           - os: windows-2019
             arch: x86
             build-group: win32-x86
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,7 +11,7 @@
       "license": "Apache-2.0",
       "dependencies": {
         "@jitsi/logger": "^2.0.0",
-        "@jitsi/robotjs": "^0.6.8",
+        "@jitsi/robotjs": "^0.6.9",
         "electron-store": "^8.0.1",
         "node-addon-api": "^4.2.0",
         "node-gyp-build": "^4.3.0",
@@ -1387,9 +1387,9 @@
       }
     },
     "node_modules/jshint": {
-      "version": "2.13.1",
-      "resolved": "https://registry.npmjs.org/jshint/-/jshint-2.13.1.tgz",
-      "integrity": "sha512-vymzfR3OysF5P774x6zYv0bD4EpH6NWRxpq54wO9mA9RuY49yb1teKSICkLx2Ryx+mfzlVVNNbTBtsRtg78t7g==",
+      "version": "2.13.4",
+      "resolved": "https://registry.npmjs.org/jshint/-/jshint-2.13.4.tgz",
+      "integrity": "sha512-HO3bosL84b2qWqI0q+kpT/OpRJwo0R4ivgmxaO848+bo10rc50SkPnrtwSFXttW0ym4np8jbJvLwk5NziB7jIw==",
       "dev": true,
       "dependencies": {
         "cli": "~1.0.0",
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -58,6 +58,7 @@ typings/
 .env
 
 build/
+prebuilds/
```
csett86, line 61: This was on purpose not on the .gitignore, as otherwise npm package will also ignore it when uploading to npm.

Thus please revert

csett86 (changes requested):
In general I agree to the approach, but from my testing the x64-only would be taken preferred to the universal build prebuild. Have you tested this on Darwin arm64 that there the universal prebuild is still preferred over the x64 one?

On a separate note: does the npm audit fix have to be part of this?

saghul:
I don't have access to an M1 mac, so I'm flying blind here, alas.

The audit fix is unrelated I just ran into it when I did npm install.

csett86:
While thinking about it, can you elaborate a bit more on the build errors that you are seeing? As the universal build for jitsi-meet-electron is internally also 2 single arch builds, and they work fine with the current prebuilds:

   • rebuilding native dependencies  dependencies=@jitsi/electron-sdk@3.0.6, @jitsi/robotjs@0.6.8 platform=darwin arch=x64
  • packaging       platform=darwin arch=x64 electron=17.1.2 appOutDir=dist/mac-universal--x64
  • downloading     url=https://github.com/electron/electron/releases/download/v17.1.2/electron-v17.1.2-darwin-x64.zip size=81 MB parts=6
  • downloaded      url=https://github.com/electron/electron/releases/download/v17.1.2/electron-v17.1.2-darwin-x64.zip duration=1.152s
  • rebuilding native dependencies  dependencies=@jitsi/electron-sdk@3.0.6, @jitsi/robotjs@0.6.8 platform=darwin arch=arm64
  • packaging       platform=darwin arch=arm64 electron=17.1.2 appOutDir=dist/mac-universal--arm64
  • downloading     url=https://github.com/electron/electron/releases/download/v17.1.2/electron-v17.1.2-darwin-arm64.zip size=84 MB parts=6
  • downloaded      url=https://github.com/electron/electron/releases/download/v17.1.2/electron-v17.1.2-darwin-arm64.zip duration=1.261s
  • packaging       platform=darwin arch=universal electron=17.1.2 appOutDir=dist/mac-universal


Separately I just locally build x64 only for jitsi-meet-electron and I cannot reproduce any build failure with the released universal build version 3.0.6.

csett86:
I tested and its at least not breaking M1 Macs, but before merging I would like to understand the issue and have the .gitignore and package-lock.json changes reverted.

saghul:
Here is the error we get:

electron build image



> and have the .gitignore and package-lock.json changes reverted.

Why though? It's a separate commit and I do want those changes in too.  Making a separate PR just to split the commit out is kinda pointless.

csett86:
Thanks for the error message, I'll check this tomorrow.

> > and have the .gitignore and package-lock.json changes reverted.
> 
> Why though? It's a separate commit and I do want those changes in too. Making a separate PR just to split the commit out is kinda pointless.

We can keep the package-lock.json and then I won't squash merge.

But the .gitignore change breaks publishing the prebuilds to npm (as everything from .gitignore is also ignored during npm package), thus please revert that, otherwise we don't have any prebuilds in the published packages after this change.

saghul:
> But the .gitignore change breaks publishing the prebuilds to npm (as everything from .gitignore is also ignored during npm package), thus please revert that, otherwise we don't have any prebuilds in the published packages after this change.

Ah, gotcha, good point!

saghul:
Updated, reverted the .gitignore change.

csett86 (approved):

csett86:
Would be interested if this release 3.0.7 now fixes your issue, as I could not reproduce the issue in any way with an x64 only build.

saghul:
I am not the one running into the problem, I did pass it to a colleague who was.

PR #213, opened 2022-03-17
refactor: guard and build native code only on windows

csett86:
the native code anyway only is required and supports windows


csett86:
Please review, @saghul

saghul (approved):
LGTM, thank you!

PR #214, opened 2022-03-17
refactor: defer loading robotjs if remotecontrol is really used

csett86:
robotjs also includes native code, defer loading of that

### DIFF
--- a/remotecontrol/render.js
+++ b/remotecontrol/render.js
@@ -1,7 +1,6 @@
 const { ipcRenderer } = require('electron');
 const os = require('os');
 const postis = require("postis");
-const robot = require("@jitsi/robotjs");
 const constants = require("./constants");
 const {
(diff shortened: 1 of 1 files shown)


saghul:
What benefit does this have? I was planning to eventually switch to using `import`, and that wouldn't work here.

csett86:
My idea was that currently under normal jitsi-meet-electron this would save loading the native code from disk that is anyway not used (as remotecontrol is disabled). This was inspired by https://www.electronjs.org/docs/latest/tutorial/performance#1-carelessly-including-modules and https://webpack.js.org/guides/lazy-loading/

csett86:
Please review, @saghul

I would cut a release after this is merged

saghul (approved):

saghul:
✅

PR #215, opened 2022-03-21
fix(ScreenSharingTracker) use a preload script

saghul:
- Enable context isolation for this window
- Add a preload script
- Disable unnecessary Node integration
- Sanitize query parameter handling

### DIFF
--- a/screensharing/screenSharingTracker.js
+++ b/screensharing/screenSharingTracker.js
@@ -1,23 +1,16 @@
-const { ipcRenderer } = require('electron');
-const querystring = require('querystring');
-
-const { SCREEN_SHARE_EVENTS_CHANNEL, SCREEN_SHARE_EVENTS } = require('./constants');
-
 const screenShareStop = document.getElementById("screen-share-marker-stop");
 const screenShareMinimize = document.getElementById("screen-share-marker-minimize");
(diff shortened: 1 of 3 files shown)


screensharing/screenSharingTracker.js
```diff
@@ -1,21 +1,20 @@
...
 const screenShareStop = document.getElementById("screen-share-marker-stop");
 const screenShareMinimize = document.getElementById("screen-share-marker-minimize");
 const sharingIdentity = document.getElementById("sharing-identity");
 
-sharingIdentity.innerHTML = querystring.parse(global.location.search)['?sharingIdentity'];
+sharingIdentity.innerText = new URL(window.location).searchParams.get('sharingIdentity');
+
+// NB: Keep in sync with contants.js
+const HIDE_TRACKER = 'hide-tracker-window';
+const STOP_SCREEN_SHARE = 'stop-screen-share';
```
horymury, line 9: any reason why we duplicated those?

horymury (changes requested):

screensharing/screenSharingTracker.js
saghul: Updated, PTAL.

PR #251, opened 2022-07-18
fix(aot) fix hiding when SS on Windows

saghul:
Fixes: https://github.com/jitsi/jitsi-meet-electron-sdk/issues/133


csett86 (approved):
I updated the electron version in the README so that it contains the fixed setContentProtection on windows, otherwise LGTM

PR #253, opened 2022-07-22
feat(aot): add capability to attach aot to any child window

quitrk:


alwaysontop/main/index.js
```diff
@@ -114,16 +119,26 @@ const showAot = () => {
...
 };
 
+/**
+ * Attaches event handlers on the undocked window
+ */
+const addUndockedWindowHandlers = () => {
+    logInfo(`adding undocked window event handlers`);
+
+    undockedWindow.on('minimize', showAot);
+    undockedWindow.on('restore', hideAot);
```
horymury, line 135: don't we need to consider blur/focus on the undocked window? eg, the undocked window always minimizes if you switch to other apps?

horymury (changes requested):

alwaysontop/main/index.js
quitrk: we dont want to potentially have both the undocked and the aot window side by side. This is similar to what other apps are doing when the meeting is a separate window than the main app.

horymury (approved):

Issue #258, opened 2022-07-25
Autopublish?


saghul:
Some time ago I created this, which we use on some of our JS packages: https://github.com/jitsi/gh-action-autopublish

The idea is simple: bbump the minor and create a new npm release when a PR gets merged.

If a PR needs to change the minor / major, that can be done as part of the PR and the script notices it and skips the bump.

Thoughts on adding it here?

quitrk:
Shouldn't we worry about releasing versions for users who get to install latest, when latest is broken?

hristoterezov:
+1 for auto publish from me!

I guess if we find out that the latest is broken we can push a hot fix or something.

csett86:
Agree as well, we just need to have the prebuildify matrix step and collection of results included so that the npm published package still contains the prebuild binaries

csett86:
Its nearly there, but failing due to protected master branch:

> remote: error: GH006: Protected branch update failed for refs/heads/master.

Is that something you can fix in the settings of the repo @saghul?

saghul:
Fixed now.

PR #264, opened 2022-08-02
ci: autopublish to npm on every commit

csett86:
Closes: #258

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,9 +2,9 @@ name: CI
 
 on:
   pull_request:
-  release:
-    types:
-      - published
+  push:
+    branches:
+      - master
 
 jobs:
   build:
@@ -41,7 +41,7 @@ jobs:
         path: prebuilds/
   publish:
     name: Publish to npm
-    if: github.event_name == 'release' && github.event.action == 'published'
+    if: github.event_name == 'push' && github.ref == 'refs/heads/master'
     runs-on: ubuntu-20.04
     needs: build
     steps:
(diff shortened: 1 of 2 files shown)


csett86:
@saghul, please review

saghul (approved):

saghul:
I removed branch protection since the that prevents autopublish from working.

csett86:
Now the npm whoami is failing:

https://github.com/jitsi/jitsi-meet-electron-sdk/runs/7669322683?check_suite_focus=true#step:6:23

Any idea how to fix that as well, @saghul?

csett86:
I already checked https://github.com/jitsi/eslint-config-jitsi/blob/master/.github/workflows/autopublish.yml and did not spot any difference

saghul:
Hum, maybe it needs a new npm auth token. I'll update it.

saghul:
Updated. Can you try again?

csett86:
I think I have an idea: #265

csett86:
Ok, still not enough, now failing on the npm install in the autopublisher, will check later

saghul:
Thanks mate!

csett86:
I did a deep dive and was unable to get it to work with the jitsi/autopublisher due to the binary dependencies.

Now switched to the original autopublisher and added the npm publish steps manually afterwards: #266 

This time I also tested it successfully with dry-run flags: https://github.com/jitsi/jitsi-meet-electron-sdk/actions/runs/2805064709

csett86:
Its working as tested, I directly proposed it to @jitsi/robotjs as well: https://github.com/jitsi/robotjs/pull/13

PR #266, opened 2022-08-05
fix(ci): autopublish v2

csett86:
jitsi/gh-action-autopublish does not work, as it tries to do npm install,
which will fail due to binary dependencies (robotjs) in the docker
container of jitsi/gh-action-autopublish


saghul (approved):

saghul:
Thanks for persevering!

PR #269, opened 2022-08-06
ref(aot) refactor aot ipc

quitrk:
@saghul @horymury 

There are 3 commits in this pr for easier review.

### DIFF
--- a/alwaysontop/index.js
+++ b/alwaysontop/index.js
@@ -1,5 +1,5 @@
 const setupAlwaysOnTopRender = require('./render');
-const { attachAlwaysOnTopToWindow, setupAlwaysOnTopMain } = require('./main');
+const { setupAlwaysOnTopMain } = require('./main');
 const { popupsConfigRegistry } = require('../popupsconfig');
 
 popupsConfigRegistry.registerPopupConfig('always-on-top', {
@@ -10,7 +10,6 @@ popupsConfigRegistry.registerPopupConfig('always-on-top', {
 });
 
 module.exports = {
-    attachAlwaysOnTopToWindow,
     setupAlwaysOnTopMain,
     setupAlwaysOnTopRender
 };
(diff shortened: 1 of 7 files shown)


alwaysontop/index.js
```diff
@@ -1,16 +1,7 @@
 const setupAlwaysOnTopRender = require('./render');
-const { attachAlwaysOnTopToWindow, setupAlwaysOnTopMain } = require('./main');
-const { popupsConfigRegistry } = require('../popupsconfig');
-
-popupsConfigRegistry.registerPopupConfig('always-on-top', {
```
quitrk, line 5: was not used

alwaysontop/index.js
```diff
@@ -1,16 +1,7 @@
...
-    matchPatterns: {
-        frameName: 'AlwaysOnTop'
-    },
-    target: 'electron'
-});
+const { setupAlwaysOnTopMain } = require('./main');
 
 module.exports = {
-    attachAlwaysOnTopToWindow,
```
quitrk, line 13: not used anymore

alwaysontop/index.js
```diff
@@ -1,5 +1,5 @@
 const setupAlwaysOnTopRender = require('./render');
-const { attachAlwaysOnTopToWindow, setupAlwaysOnTopMain } = require('./main');
```
saghul, line 2: This was only used in the intermediate undocked implementation, right? Else, we'd have to bump the major since it's an API breakage.

alwaysontop/index.js
saghul: 🚀

saghul (approved):

alwaysontop/index.js
quitrk: that's correct

PR #280, opened 2022-08-26
chore(build) prevent accidental pushes to master

saghul:
The auto-release bot should still work since it doesn't install the
pre-commit hook.

Sample run:

```
$ git push --dry-run 
You can't commit directly to master branch
husky - pre-push hook exited with code 1 (error)
error: failed to push some refs to 'github.com:jitsi/jitsi-meet-electron-utils.git'
```

### DIFF
--- a/.husky/pre-push
+++ b/.husky/pre-push
@@ -1,0 +1,10 @@
+#!/usr/bin/env sh
+. "$(dirname -- "$0")/_/husky.sh"
+
+branch="$(git rev-parse --abbrev-ref HEAD)"
+
+if [ "$branch" = "master" ]; then
+  echo "You can't commit directly to master branch"
+  exit 1
+fi
+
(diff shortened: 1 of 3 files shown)


horymury (approved):

saghul:
@csett86 do you reckon this would work for you?

csett86:
Are you sure that the checkout action (https://github.com/jitsi/jitsi-meet-electron-sdk/blob/master/.github/workflows/ci.yml#L48) does not install the precommit hook?

i could not find anything pointing to that…

saghul:
Reasonably sure, since I added that step myself to my auto-publish plugin, which is what created trouble, so I think it's safe to say checking it out doesn't npm i.

saghul:
You can also check in any of the runs: https://github.com/jitsi/jitsi-meet-electron-sdk/runs/8034378226?check_suite_focus=true

csett86 (approved):
Then lets go for this

csett86:
Fails in a different place, unfortunately: https://github.com/jitsi/jitsi-meet-electron-sdk/runs/8035799191?check_suite_focus=true

saghul:
FML, `npm publish` runs the `prepare` step too... Let me check if we can skip it...

PR #281, opened 2022-08-26
feat(aot-prejoin) Enable aot on meeting prejoin

horymury:


alwaysontop/render/index.js
```diff
@@ -37,11 +37,14 @@ class AlwaysOnTop extends EventEmitter {
      * Creates new instance.
      *
      * @param {JitsiIFrameApi} api - the Jitsi Meet iframe api object.
+     * @param {boolean} showOnPrejoin - whether AOT should show on prejoin screen.
      */
-    constructor(api) {
+    constructor(api, showOnPrejoin) {
```
saghul, line 42: Can you please add an `options` object instead, so we can add more things if / when we need to?

alwaysontop/render/index.js
```diff
@@ -79,6 +90,12 @@ class AlwaysOnTop extends EventEmitter {
 
     _onConferenceJoined() {
         logInfo('on conference joined');
+        if (this._joined) {
```
saghul, line 93: Why is this logic necessary?

alwaysontop/render/index.js
```diff
@@ -293,9 +318,10 @@ class AlwaysOnTop extends EventEmitter {
 *
 * @param {JitsiIFrameApi} api - the Jitsi Meet iframe api object.
 * @param {Logger} loggerTransports - external loggers
+* @param {boolean} showOnPrejoin - whether AOT should show on prejoin screen
 */
-module.exports = (api, loggerTransports) => {
+module.exports = (api, loggerTransports, showOnPrejoin) => {
```
saghul, line 323: let's do an options object before this gets out of hand...

alwaysontop/render/index.js
horymury: cause _onConferenceJoined is called twice when showOnPrejoin, as prejoinScreenLoaded and videoConferenceJoined both trigger it.

alwaysontop/render/index.js
horymury: with keeping backwards compat? it would be a bit tricky, as I don t see an easy way to discern whether the caller used the new or old signature for it.

horymury: something along the lines `module.exports = (options) { if (options.hasOwnProperty('api'))...` ? as api would be a mandatory prop.

saghul: No I mean add the new option inside an options object...

alwaysontop/render/index.js
saghul: Also breakoput rooms, but what is the issue?

alwaysontop/render/index.js
horymury: oh got it

alwaysontop/render/index.js
horymury: I thought to avoid unexpected issues by it being called twice, but I'll test without it as it might be fine without the check.

saghul: 👍

horymury: @saghul retested without the flag and it's not working as expected, it actually revealed an existing issue on AOT (without my changes). Seems like ipcRenderer event handler is not cleaned properly on leave if `ipcRenderer.on(...` is called multiple times.

quitrk: good catch

alwaysontop/render/index.js
```diff
@@ -292,10 +320,13 @@ class AlwaysOnTop extends EventEmitter {
 * window which displays Jitsi Meet.
 *
 * @param {JitsiIFrameApi} api - the Jitsi Meet iframe api object.
-* @param {Logger} loggerTransports - external loggers
+* @param {Logger} loggerTransports - external loggers.
+* @param {Object} options - AOT options.
 */
-module.exports = (api, loggerTransports) => {
+module.exports = (api, loggerTransports, { showOnPrejoin = false } = {}) => {
```
quitrk, line 326: we never did anything with `loggerTransports`. Can't we just remove it?

alwaysontop/render/index.js
```diff
@@ -292,10 +320,13 @@ class AlwaysOnTop extends EventEmitter {
...
     setLogger(loggerTransports);
 
-    return new AlwaysOnTop(api, loggerTransports);
+    return new AlwaysOnTop(api, {
+        showOnPrejoin :showOnPrejoin && typeof api._getPrejoinVideo === 'function'
```
quitrk, line 330: nit: wrong whitespace placement

horymury: oops, fixed.

quitrk (approved):

saghul (approved):

PR #282, opened 2022-08-26
fix(build): Disable autoinstall and document husky instead 

saghul:
Avoid the prepare step, which also runs when publishing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -172,6 +172,10 @@ For examples of installation and usage checkout the [Jitsi Meet Electron](https:
 
 ## Development
 
+Enable husky to avoid accidental pushes to the main branch:
+
+    npx husky install
+
 To rebuild the native code, use:
 
     npx node-gyp rebuild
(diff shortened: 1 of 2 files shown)


csett86:
Unfortunately this also does not work, as then every install of the package (also as a normal dependency eg as part of jitsi-meet-electron) would case husky install to be run, which then fails as well:

```
(base) chris@krypton tmp % npm install --save 'jitsi/jitsi-meet-electron-sdk#husky-postinstall'
npm ERR! code 127
npm ERR! path /Users/chris/workspace/node_modules/@jitsi/electron-sdk
npm ERR! command failed
npm ERR! command sh -c husky install
npm ERR! sh: husky: command not found
```

So for now I suggest that we document that for developers but not force husky as I dont see a way how to have husky only available locally but not as part of any CI chain, based on the experience here and https://docs.npmjs.com/cli/v6/using-npm/scripts

csett86 (approved):

Issue #279, opened 2022-08-26
Auto-Release: create a GH release too


saghul:
Title says it all.

PR #305, opened 2022-11-20
WIP: use release-it to also have github releases

csett86:
Closes: #279


csett86:
It would be nice, but no time to fix this to have it working, thus closing. Also it would complicate the CI even further

PR #310, opened 2022-12-10
popupconfig: adjust to new-window event replacement

csett86:
See https://www.electronjs.org/docs/latest/breaking-changes#removed-webcontents-new-window-event

### DIFF
--- a/popupsconfig/main.js
+++ b/popupsconfig/main.js
@@ -1,5 +1,3 @@
-const electron = require('electron');
-const { BrowserWindow } = electron;
 const popupsConfigRegistry = require('./PopupsConfigRegistry');
 const { testMatchPatterns } = require('./functions');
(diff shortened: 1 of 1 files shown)


csett86:
@saghul please review if you have time

saghul:
I'd like @quitrk to take a look since we need to integrate with other window handlers in 8x8 apps and we need to keep that working.

csett86:
@quitrk can you please take a look? Thanks in advance!

csett86:
@quitrk as the new-window event is removed since Electron 22 (and we are now at 25), and the current popupsconfig code still relies on it, can you please take a look if this replacement PR looks good from your perspective?

quitrk:
@csett86 looks good to me

quitrk (approved):

csett86:
Thank you @quitrk

PR #315, opened 2022-12-22
ci: update to latest and limit concurrency on publish job

csett86:


csett86:
@saghul please review if you have time

saghul (approved):

PR #322, opened 2023-01-23
Drop support for WifiStats

saghul:


hristoterezov:
Why are we dropping wifi stats?

hristoterezov (changes requested):
Please add some details? Is it not working anymore? We don't need it anymore?

saghul:
We don't use it / need it anymore indeed.

hristoterezov (approved):
Let's remove it from the other projects too!

PR #333, opened 2023-03-22
feat(screenshare) - add web security fix for electron

dudumanbogdan:
- add web security fix for electron using external_api 

First solution used was:
`  window.navigator.mediaDevices.getDisplayMedia()` in `jitsi-meet` and in `jitsi-meet-electron` to listen in main process on `session.defaultSession.setDisplayMediaRequestHandler(async (request, callback) => {`
- than we could build a new picker logic in jitsi-meet-electron and when a thumbnail was selected to call the `callback` with the video item based on the source Id selected
- the problem with this solution is that we couldn't have anymore audio screen share. In `lib-jitsi-meet` ScreenObtainer we get from main process the stream selected, we don't have anymore `navigator.mediaDevices.getUserMedia`


hristoterezov:
@gabiborlea could you please add some context on what we try to fix with this change?

saghul (approved):

gabiborlea:
> @gabiborlea could you please add some context on what we try to fix with this change?
It is intended to fix the security concern that is added with the `JitsiMeetElectron` object being added to the iframe window object. With this approach, the external api will be used to provide the stream for the screens. There is also backward compatible. The jitsi meet pr is here: https://github.com/jitsi/jitsi-meet/pull/13096.

saghul (approved):

PR #335, opened 2023-04-19
fix(remote-control) Move robotjs import

horymury:
This is in order to avoid 
image

on Electron apps using electron v23.0.0 and newer

### DIFF
--- a/remotecontrol/render.js
+++ b/remotecontrol/render.js
@@ -2,6 +2,8 @@ const { ipcRenderer } = require('electron');
 const os = require('os');
 const postis = require("postis");
 const constants = require("./constants");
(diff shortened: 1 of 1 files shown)


damencho (approved):

horymury:
> 

Yep they changed something starting electron 23.0.0 and we would get the mentioned error unfortunately.

csett86:
The upstream issue is https://github.com/electron/electron/issues/37404, this is affecting other modules apart from robotjs as well

csett86:
@horymury and just a fyi, Wayland screensharing was also completely broken (as in segfault every time when trying) with electron 22 or later, see https://github.com/jitsi/jitsi-meet-electron/issues/829

PR #354, opened 2023-10-18
fix(screensharing): add screensharing supported request

gabiborlea:


screensharing/render.js
```diff
@@ -31,66 +31,109 @@ class ScreenShareRenderHook {
         this._iframe.addEventListener('load', this._onIframeApiLoad);
     }
 
+    // this is a rewrite of the sendRequest method on the transport
```
saghul, line 34: Ugh, so we don't currently respond to requests we don't know about? That's bad...

saghul: I think this is too much. I know it's more brittle, but doesn't matter since there is a fallback. Let's use a timeout please. Given this price is only paid once, it shouldn't be a huge deal...

saghul: Also, 10 secodns is waaay to much, we should get an answer within 2 secods I think.

screensharing/render.js
```diff
@@ -31,66 +31,109 @@ class ScreenShareRenderHook {
...
-                                    types: data.options.desktopSharingSources
-                                }
-                            }
-                        });
-                    })
-                    .catch((error) => {
-                        this._api.executeCommand('_requestDesktopSourcesResult', { error: { data: error  }  });
-                    });
+    async _onIframeApiLoad() {
+        this._iframe.contentWindow.JitsiMeetElectron = {
```
saghul, line 89: Add a TODO: delete this after 2 releases, please.

saghul (changes requested):

screensharing/render.js
gabiborlea: A timeout alone would not help, we have to abort the promise

gabiborlea: This is why I wrote this function so that I can pass the controller signal in the promise constructor

saghul (approved):
LGTM! @hristoterezov can you PTAL too?

screensharing/render.js
```diff
@@ -31,66 +31,70 @@ class ScreenShareRenderHook {
...
+        return new Promise((resolve) => {
+            this._api._isNewElectronScreensharingSupported()
+                .then(response => {
+                    if (response.error){
+                        resolve(false);
+                    }
+                    resolve(response);
+                })
+                .catch(() => resolve(false));
+            setTimeout(() => resolve(false), 2000);
```
hristoterezov, line 46: shouldn't we cancel the timeout if previous call resolves?

screensharing/render.js
```diff
@@ -31,66 +31,70 @@ class ScreenShareRenderHook {
...
+        const isNewElectronScreensharingSupported = await this._isNewElectronScreensharingSupported();
+
+        if (isNewElectronScreensharingSupported) {
+            this._iframe.contentWindow.JitsiMeetElectron = undefined;
+            this._api.on('_requestDesktopSources', async (request, callback) => {
+                const { options } = request;
+
+                ipcRenderer.invoke(SCREEN_SHARE_GET_SOURCES, options)
+                    .then((sources)=> sources.map(item => {item.thumbnail.dataUrl = item.thumbnail.toDataURL(); return item;}))
+                    .then((sources) => callback({sources}))
```
hristoterezov, line 91: Wouldn't it be simpler if we have 1 `then`?

saghul (approved):

PR #369, opened 2024-01-16
fix(screen-sharing) fix depredcated ss hook

saghul:
We may fail to access the iframe's contentWindow, but we can ignore that since the new hook doesn't require it.

Fixes: https://github.com/jitsi/jitsi-meet-electron-sdk/issues/368
Fixes: https://github.com/jitsi/jitsi-meet-electron-sdk/issues/325


csett86 (approved):
Looks good to me

PR #372, opened 2024-02-01
fix: Make AOT window be on top of fullscreen apps on MacOS 14

miro-ku:
Hey there. I've noticed that AOT window doesn't stay on top of full-screen apps on macOS Sonoma (v14.2.1).
Not sure if it's the same on lower versions and/or Windows, but setting `type: 'panel'` for created `BrowserWindow` resolves the issue for me

### DIFF
--- a/alwaysontop/main/config.js
+++ b/alwaysontop/main/config.js
@@ -1,6 +1,7 @@
 const { SIZE } = require("../constants");
 
 module.exports = {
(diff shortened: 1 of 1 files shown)


csett86:
From documentation, type: 'panel' is a macOS only-option, so I'd rather not set that in general on the AOT for all platforms

PR #387, opened 2024-04-21
ci: update to v4 actions

csett86:
the upload action no longer allows "appending" to the same name, so merge on download as described https://github.com/actions/download-artifact?tab=readme-ov-file#download-multiple-filtered-artifacts-to-the-same-directory


csett86:
Contains #366 and #367

saghul (approved):

PR #390, opened 2024-04-21
ci: fix prebuild download on release

csett86:


csett86:
I should have read https://github.com/actions/download-artifact?tab=readme-ov-file#download-multiple-filtered-artifacts-to-the-same-directory properly, name is replaced with pattern, and path stays as is. I #387 I accidentally replaced path with pattern...

saghul (approved):

PR #434, opened 2025-02-06
feat(screensharing) add support for getDisplayMedia

saghul:


csett86:
Thank you for picking this up! Just a minor question: Is this really a breaking change? Otherwise I would leave the version at 6.*, and just do a minor bump, eg. to 6.1.0

saghul:
You are right. When I started it looked like it needed to be a breaking change, but as I got deeper things cleared out so I can indeed just bump the minor.

package-lock.json
```diff
@@ -1,12 +1,12 @@
 {
   "name": "@jitsi/electron-sdk",
-  "version": "6.0.50",
+  "version": "7.0.0",
```
hristoterezov, line 3: This needs to be updated.

screensharing/main.js
```diff
@@ -85,6 +109,44 @@ class ScreenShareMainHook {
...
+                const options = {
+                    types: [ 'screen', 'window' ],
+                    thumbnailSize: {
+                        width: 0,
+                        height: 0
+                    },
+                    fetchWindowIcons: false
+                };
+
+                desktopCapturer.getSources(options).then(sources => {
```
hristoterezov, line 133: Am I correct that this should be the second `desktopCapturer.getSources(` call for the whole process - one from the desktop picker and this one?

Isn't it possible to avoid this?

saghul: Yes, this is the 2nd one. The problem is we need the object, not the ID. I can try to cache them on the 1st call, but note this 2nd one doesn't fetch icons or thumbnails so it's much cheaper to run.

saghul:
@hristoterezov Fixed the version mismatch. PTAL!

screensharing/main.js
saghul: I added a cache but still kept the fallback behavior in case they go out of sync. The window is very narrow, but let's err on the safe side.

saghul:
On a second thought, I think we should bump the major to 7. https://github.com/jitsi/jitsi-meet-electron-sdk/pull/430 should've probably done that, and now seems like a good time to do it.

The screen-sharing implementation has gone through quite a bit:

- 1st it relied on injecting stuff into the iframe -> bad!
- 2nd we implemented the "new" flow
- 3rd, now, we are adding (hopefully) The Final Way To Do It: getDisplayMedia

So, I propose we bump to 7 now, since we got rid of option 1 in Jitsi Meet, and this signals that.

Then, in 6 months - 1 year bump it to 8 and drop option 2, leaving gDM as the one and only option and removing all vestiges of the previous ones from JM and LJM.

Thoughts?

csett86:
I agree that the legacy flow removal should have been a major bump, so lets do that (retrospectively) now

Thank you for giving this more thought, @saghul

screensharing/main.js
```diff
@@ -85,6 +105,62 @@ class ScreenShareMainHook {
...
+                this._gdmData = null;
+
+                if (!data.streamId) {
+                    callback(null);
+
+                    break;
+                }
+
+                // The API requires a `DesktopCapturerSource` object. Find the matching one.
+                const constraints = {
```
hristoterezov, line 120: So according to this https://www.electronjs.org/docs/latest/api/session#sessetdisplaymediarequesthandlerhandler-opts we need to provide only id and name. Doesn't this work? 

If it works everything should be way more simpler.

saghul: It doesn't. It needs the object.

hristoterezov: Have you tested it? I tested it on MacOS and it works with passing only: 
```
{ video: {id,name} }
```
as described in the docs!

saghul: I only passed the id... maybe it needs the name too? I'll test!

hristoterezov:
@saghul I've also tested on Linux (Ubuntu 24.04) and the whole app is crashing after the callback is called!

Have you tested on Linux? I'm wondering if it some misconfiguration on my side because I can't find many complaints online...

Otherwise it is crashing with:
```
'loop->recurse > 0' failed at ../src/pipewire/thread-loop.c:426 pw_thread_loop_wait()
[1] /home/hristo/Documents/workspace/jitsi-meet-electron/node_modules/electron/dist/electron exited with signal SIGSEGV
``` for me!

**Comment by saghul** [comment]
PipeWire is known to cause crashes, or at least it has in the past.

Not sure how this could cause it but I'll see if I can repro.

**Comment by hristoterezov** [comment]
Did a second test and what actually happens is:
 - Press the screen sharing button 
 - Our desktop picker is displayed and also the native linux SS picker window pops up
At this point I can't go beyond this window, after a selection I see the error in the console and the native linux SS picker window pops up again and again. At the end if I decide to close it the whole app crashes...

**Comment by saghul** [comment]
This is a known bug with pipewire. Not new.

IIRC you need to be very quick selecting it and it will work.

Setting the system picker to true should avoid it, and was next on my list.

File: screensharing/main.js (hunk 3 above)
Comment (hristoterezov) [reply screensharing/main.js:120]: Yep. I was passing the name too and it worked

Comment (saghul) [reply screensharing/main.js:119]: Alright, I had to modify Jitsi Meet to pass around the entire object (why not!) rather than the id alone and it works indeed.

**Comment by saghul** [comment]
@hristoterezov I was wrong about the system picker, it currently is only implemented for macOS >= 15 and has its own quirks, so I'm leaving it disabled.

As for SS on Linux, see: https://github.com/jitsi/jitsi-meet-electron?tab=readme-ov-file#gnulinux

I'll give it a test though.

**Comment by csett86** [comment]
@saghul Can we maybe use this opportunity to add the trick mentioned here: https://github.com/jitsi/jitsi-meet-electron/issues/963#issuecomment-2114765249

**Comment by saghul** [comment]
Good news! I implemented the Wayland workaround and it works! For some reason the PipeWire picker is displayed twice 🤷  but once you pick what you want to share it works without displaying our dialog and weird warnings. We don't control the portal, that is displayed as a result to calling `desktopCapturer.getSources`.

PipeWire crashing has nothing to do with this PR, alas. I upgraded my system and didn't get crashes.

(I've waited for this moment for a long time) I USED ARCH BY THE WAY.

Things are, of course, still working fine in X11.

File: screensharing/main.js [hunk 4]
```diff
@@ -26,15 +32,59 @@ class ScreenShareMainHook {
...
+        );
+
         // Listen for events coming in from the main render window and the screen share tracker window.
-        electron.ipcMain.on(SCREEN_SHARE_EVENTS_CHANNEL, this._onScreenSharingEvent);
-        electron.ipcMain.handle(SCREEN_SHARE_GET_SOURCES, this._onGetSourcesInvoke);
+        ipcMain.on(SCREEN_SHARE_EVENTS_CHANNEL, this._onScreenSharingEvent);
+        ipcMain.handle(SCREEN_SHARE_GET_SOURCES, async (_event, opts) => {
+            const sources = await desktopCapturer.getSources(opts);
+
+            this._cachedSources = sources;
```
hristoterezov, line 84: no need of this anymore

screensharing/main.js
```diff
@@ -85,6 +123,32 @@ class ScreenShareMainHook {
...
+                    callback(null);
+
+                    break;
+                }
+
+                const constraints = {
+                    video: data.source
+                };
+
+                // Setting `audio` to `undefined` throws an exception.
```
hristoterezov, line 141: what does this mean?

I guess we still don't set it in certain cases which will be equivalent to setting it to undefined...

screensharing/main.js
saghul: Ops, yeah, forgot to remove it!

screensharing/main.js
saghul: It's only supported if it's a string. If absent it's ok. An API quirk.

hristoterezov: Nice

hristoterezov (approved):
Great work!

I love the change for the wayland! 

Retested on my linux machine and Now everything seems to work fine! I think before it was combination between me picking to share the whole screen and also I believe I somehow ended up using not the correct version of the sdk....

Issue #461, opened 2025-09-22
Sometimes SDK throws error when starting screen share


fearistoff:
When user starts screen sharing sometimes SDK throws error `TypeError: Cannot destructure property 'callback' of 'this._gdmData' as it is null`.

<img width="918" height="514" alt="Image" src="https://github.com/user-attachments/assets/b37670fb-9bff-4a86-ba81-cc0774ae937f" />

Error was reproduced with version 7.0.8 of SDK. Server runs version [2.0.10431](https://github.com/jitsi/jitsi-meet/releases/tag/stable%2Fjitsi-meet_10431) of Jitsi. It throws rapidly - about 1% chance.

User just starts screen sharing on Windows machine (mostly Windows 10) and error appear and it keep showing unless user close application and open again.

PR #462, opened 2025-09-22
Add _gdmData check for null

fearistoff:
Fixes issue #461


saghul:
That doesn't fix the real problem, which is: why was the DO_GDM action requested when there was no gDM data stored? What SDK version are you using?

fearistoff:
Error throws rapidly - about 1% chance.

User just starts screen sharing on Windows machine (mostly Windows 10).

saghul:
Can you try and debug the sequence of events? DO_GDM is only sent when the user picks a surface, is it possible that it gets called twice?

PR #501, opened 2026-05-30
replace @jitsi/logger with native console logs

csett86:


saghul:
I think we cannot do this because embedding apps use the transports parameters.

@hristoterezov can confirm.

hristoterezov:
> I think we cannot do this because embedding apps use the transports parameters.
> 
> @hristoterezov can confirm.

Yes! This is correct!  We are using all these. 

I'll also appreciate if we keep the logs for debugging purposes for the time being. We have a few reports of PiP not showing when it is supposed to show which we are trying to fix.

what is the reason for the change? Just trying to simplify and clean?

csett86:
@hristoterezov @saghul Yes the reason was purely to simplify and clean, and it initially I got the idea when trying to make the SDK better embeddable for esbuild etc in https://github.com/jitsi/jitsi-meet-electron-sdk/pull/500

But if this is currently used and helpful, lets skip this for now

hristoterezov:
@csett86 thanks!

PR #506, opened 2026-07-01
ci: build with node 24

csett86:
Move to current LTS


csett86:
Nevermind, 

> Unable to find Node version '24' for platform win32 and architecture x86.

PR #507, opened 2026-07-14
BREAKING CHANGE: Enable contextIsolation

saghul:
(See individual commits)


pip/bridge.js
```diff
@@ -0,0 +1,27 @@
...
+    return {
+        /**
+         * Requests the main process to trigger picture-in-picture for the given
+         * frame. Non-string frame names are ignored (the main process guards
+         * against a falsy frame name anyway).
+         *
+         * @param {string} frameName - The name of the iframe requesting PiP.
+         * @returns {void}
+         */
+        request: frameName => {
```
hristoterezov, line 21: Did you use this syntax on purpose? 

Won't `request(frameName) { ...` work?

saghul: This is an object literal, not a class so that syntax works.

hristoterezov: you mean `request(frameName) { ...` won't work?

saghul: Yes, that won't work.

README.md
```diff
@@ -2,203 +2,306 @@
...
 
-On every push to main branch, the .github/workflows/ci.yml will create a new version and publish to npm.
-
-If a major or minor release is required, use respective key words in the commit message, see https://github.com/phips28/gh-action-bump-version#workflow
+On every push to the `master` branch, `.github/workflows/ci.yml` creates a new version and
+publishes to npm. For a major or minor release, use the respective keywords in the commit
+message — see the
+[gh-action-bump-version workflow](https://github.com/phips28/gh-action-bump-version#workflow).
+</content>
+</invoke>
```
hristoterezov: Leftover tool-call markup — these two lines (`</content>`, `</invoke>`) made it into the committed file and need to be deleted.

remotecontrol/main.js
```diff
@@ -1,56 +1,207 @@
...
-        event.returnValue = this._getDisplay(sourceId);
+    _handleEvent(event, data) {
+        // Ignore events until a display has been associated with the session.
+        if (!this._display || !data) {
+            return;
+        }
+
+        const robot = this._robot;
+
+        switch (data.type) {
```
hristoterezov: Now that these robotjs calls run in the main process, an exception thrown here escapes the `ipcMain.on` listener as an uncaught exception **in the main process** — which by default takes the whole app down. robotjs throws on invalid input (e.g. `keyToggle` throws `Invalid key code specified.` for an unrecognized `key`; unknown `modifiers` throw too), and the preload sanitizer deliberately lets any string through for `key`/`modifiers`. Pre-v10 the same throw was contained to the renderer page.

Suggest wrapping this switch in a try/catch that logs and drops the event, so a malformed (or hostile — this is exactly the compromised-main-world scenario the sanitizers exist for) event can't crash the app.

remotecontrol/bridge.js
```diff
@@ -0,0 +1,92 @@
...
+    if (typeof event.x === 'number') {
+        sanitized.x = event.x;
+    }
+    if (typeof event.y === 'number') {
+        sanitized.y = event.y;
+    }
+    if ('button' in event) {
+        sanitized.button = event.button;
+    }
+    if (typeof event.key === 'string') {
```
hristoterezov: Since this sanitizer is the defense-in-depth boundary, consider tightening it further:
- `Number.isFinite(event.x)` rather than `typeof` — `NaN`/`Infinity` are `'number'` and flow straight into `robot.moveMouse()`.
- Validate `key`, `modifiers` and `button` against known whitelists (`MOUSE_BUTTONS`, the modifier names robotjs accepts) instead of accepting arbitrary strings — anything unrecognized currently reaches robotjs and makes it throw in the main process (see comment on `remotecontrol/main.js`).

screensharing/bridge.js
```diff
@@ -0,0 +1,132 @@
...
+        return null;
+    }
+
+    const sanitized = { name };
+
+    if ('requestId' in data) {
+        sanitized.requestId = data.requestId;
+    }
+    if ('source' in data) {
+        sanitized.source = data.source;
```
hristoterezov: `source` is forwarded verbatim and ends up in `callback({ video: data.source })` inside the getDisplayMedia handler in main — a malformed value from a compromised main world would throw inside that `ipcMain.on` listener (uncaught, in the main process). Since this sanitizer already whitelists fields, it may be worth shaping `source` down to `{ id, name }` strings too (that's all the gdm callback needs).

renderer/bridge.js
```diff
@@ -0,0 +1,41 @@
...
+
+    if (!sdk) {
+        throw new Error(
+            '[@jitsi/electron-sdk] window.jitsiElectronSDK is not available. Ensure the app\'s preload '
+            + 'script imports \'@jitsi/electron-sdk/preload\' and that the window is created with '
+            + 'contextIsolation enabled.'
+        );
+    }
+
+    if (sdk.apiVersion !== BRIDGE_API_VERSION) {
```
hristoterezov: what is the scenario where those 2 would be different?

renderer/bridge.js
```diff
@@ -0,0 +1,41 @@
...
+
+/**
+ * Retrieves a feature fragment from the `window.jitsiElectronSDK` bridge,
+ * verifying that the SDK preload is installed and speaks a compatible protocol.
+ *
+ * @param {string} feature - The bridge fragment name (e.g. 'screenSharing').
+ * @returns {Object} The requested bridge fragment.
+ * @throws {Error} If the preload is missing, out of date, or lacks the feature.
+ */
+function getBridge(feature) {
```
hristoterezov: can't we remove this and just access `window?.jitsiElectronSDK?.feature`?

screensharing/bridge.js
saghul: A compromised main means RCE, so there is nothing to protect against, we are toast.

README.md
saghul: lol

helpers/ipcRouter.js
```diff
@@ -0,0 +1,193 @@
...
+    let entry = invokeChannels.get(channel);
+
+    if (!entry) {
+        entry = { routes: [] };
+        invokeChannels.set(channel, entry);
+        ipcMain.handle(channel, (event, ...args) => {
+            const matched = routeFor(entry.routes, event);
+
+            if (!matched) {
+                return { error: 'Error: rejected IPC from an untrusted sender' };
```
hristoterezov: Returning `{ error }` assumes every invoke channel's contract is "object with optional `error`", but the callers differ: for `SCREEN_SHARE_GET_SOURCES` the preload immediately does `sources.map(serializeSource)`, so an untrusted sender gets an opaque `TypeError: sources.map is not a function` instead of a clean failure; for the power monitor the object carries no `id`/`type: 'response'`, so the iframe's pending query would never settle.

Throwing an `Error` here would make `ipcRenderer.invoke` reject uniformly — every caller already handles rejection.

powermonitor/main.js
```diff
@@ -1,106 +1,153 @@
...
+
+        // Queries are request/response: routing by sender sends the reply back
+        // to the invoking window automatically, so there is no module-level
+        // browserWindow and no webContents.send for responses.
+        addInvokeRoute(POWER_MONITOR_QUERIES_CHANNEL, {
+            owner: this._webContents,
+            handler: this._handleQuery
+        });
+
+        app.whenReady().then(() => this._attachEvents());
```
hristoterezov, line 54: Small race: if the window is closed before `app.whenReady()` resolves, `_attachEvents()` runs *after* `cleanup()` — the `powerMonitor` listeners get attached with nothing left to remove them (the instance is already out of `_instances`). A `_disposed` flag checked here (or at the top of `_attachEvents`) closes it.

powermonitor/renderer.js
```diff
@@ -88,12 +93,13 @@ module.exports = function setupPowerMonitorRender(api) {
...
-            ipcRenderer.on(POWER_MONITOR_QUERIES_CHANNEL, queriesChannelListener);
-            ipcRenderer.on(POWER_MONITOR_EVENTS_CHANNEL, eventsChannelListener);
+            _unsubscribeEvents = _bridge.onEvent(eventsChannelListener);
             _channel.listen('message', message => {
                 const { name } = message.data;
                 if(name === POWER_MONITOR_MESSAGE_NAME) {
-                    ipcRenderer.send(POWER_MONITOR_QUERIES_CHANNEL, message);
+                    _bridge.query(message)
+                        .then(response => _sendMessage(response))
+                        .catch(() => { /* window tearing down; drop the response */ });
```
hristoterezov: If the invoke rejects for any reason other than teardown (e.g. `getSystemIdleState` throwing on a bad `idleThreshold`, which `sanitizeQuery` passes through unvalidated), the iframe's pending query never gets a response and hangs. Consider replying in the catch with `{ id: message.id, error: String(error), type: 'response' }` — the `_channel` guard in `_sendMessage` still covers the genuine teardown case.

screensharing/bridge.js
hristoterezov: To clarify — "compromised **main world**" meant the page's JS context in the *renderer* (the world `exposeInMainWorld` exposes to), not the main process. Same threat model as the other sanitizers in this PR — `remotecontrol/bridge.js` uses the same wording.

The scenario: hostile/XSS'd page code calls `sendEvent({ name: DO_GDM, requestId, source: <garbage> })`. `source` crosses this sanitizer untouched and main runs `callback({ video: data.source })`, which can throw inside the `ipcMain.on` listener — an uncaught exception in the main process, i.e. a renderer-triggered app crash (DoS), not RCE. It's the same class of issue as the robotjs throw that 53ef560 hardened; shaping `source` down to `{ id, name }` strings closes it the same way.

hristoterezov (changes requested):
Awesome work! Posting some small improvement suggestions!

powermonitor/main.js
saghul: Windows cannot be open before the app is ready AFAIK so this sounds like an impossible situation.

hristoterezov (approved):
Awesome work!

PR #508, opened 2026-08-03
feat(remotecontrol): gate session start on user consent in the main process

saghul:
Require explicit consent before a session starts, collected in the main process:
a native modal message box parented to the meeting window, which web content can
neither render, click nor dismiss. This holds even if both the iframe and the
renderer hosting it are fully compromised.

_handleStart becomes async (ipcMain.handle already supports promise-returning
handlers and the renderer already awaits bridge.start, so the protocol is
unchanged) and refuses to resolve a display until consent is granted. A pending
flag rejects stacked requests, a throwing consent callback counts as a denial,
and a window destroyed while the dialog was up is handled.

Embedders can pass requestConsent to supply their own UX - a localized dialog,
for instance - as long as it is not renderable or dismissable by web content, or
requestConsent: false to disable the gate, which is only defensible when a start
request cannot originate from untrusted web content (a kiosk or appliance loading
a single deployment the embedder controls). A warning is logged when it is off.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -58,8 +58,9 @@ Communication between processes uses Electron's IPC (ipcMain/ipcRenderer) and po
 
 #### Remote Control (`remotecontrol/`)
 Enables remote desktop control during Jitsi Meet sessions:
-- `RemoteControlMain`: Runs in main process, handles display metrics and IPC
-- `RemoteControl`: Runs in renderer, uses robotjs to execute mouse/keyboard events from remote participants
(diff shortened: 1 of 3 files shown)


hristoterezov (approved):
LGTM! But we have similar dialog coming from jitsi-meet. Do you plan to keep both? Should this one replace the one from jitsi-meet? If yes maybe we should try to render the name of the requestor.

saghul:
Discussed elsewhere: this dialog is relevant only for untrusted origins. We will consider removing the builtin Jitsi Meet one after this SDK has been deployed.

